PR #22, opened 2015-03-31
Merge marked testing into the dev branch

jameshiebert:
I have created a branch of `pdp` that has all tests with external dependencies marked (e.g. bulk NetCDF that are not appropriate for revision control), which allows for running a subset of them outside of a PCIC environment (e.g. TravisCI or AWS).

I *believe* that this branch is ready to be merged into `dev`, but it would be good to have a second set of eyes on it just to make sure. There are a couple changes to the overall application structure, so I don't want to shoot us in the foot, so close to 2.3.0 deployment.

The changes include:

* Tests requiring the CRMP database are marked "crmpdb" and tests requiring external bulk data are marked "bulk_data". One can exclude this set of tests by running `py.test -m "not crmpdb and not bulk_data" tests` (this is now the default under TravisCI).
* I have moved all of the top-level WSGI application construction to *functions* in the `pdp.main`. Then I have retained *instantiation* of those functions in the `pdp.wsgi` module. This allows us to import `pdp.main` in testing modes and instantiate the applications without analytics or authentication wrappers. But the instantiated versions can still be passed over to `flask` or `gunicorn` or other application servers that require an instantiated WSGI application.
* I have changed the configuration reading code from looking for `config.yaml` as a package resource to reading the `PDP_CONFIG` environment variable. This allows us to install pdp prior to configuration and/or maintain multiple config sets (with the package or otherwise) which can be executed independently. This requires an extra deploy step, so you need to be aware of it.

I've deployed this branch to a VirtualBox instance and all 50 of 50 tests pass (after the usual tweaks to `config.yaml`), so I *think* it's good to go. But take a look, and if it looks good, just merge the pull request.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,10 +25,6 @@ def session_dir(request):
     request.addfinalizer(lambda: dirname.remove(rec=1, ignore_errors=True))
     return str(dirname)
 
-@pytest.fixture(scope="module")
-def static_url_space():
-    return frontend
-
 @pytest.fixture(scope="function")
 def raster_pydap():
     from pdp.portals.bcsd_downscale_canada import portal
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,10 +14,13 @@ import pytest
 from webob.request import Request
 from beaker.middleware import SessionMiddleware
 
-from pdp.wsgi import frontend
-
 @pytest.fixture(scope="module")
 def session_dir(request):
+    '''For testing we should manage the session directory ourselves so that:
+       a) we don't stomp on existing session directories
+       b) we don't require run time configuration
+       c) we ensure that everything gets properly cleaned up after the test run
+    '''
     dirname = py.path.local(mkdtemp())
     request.addfinalizer(lambda: dirname.remove(rec=1, ignore_errors=True))
     return str(dirname)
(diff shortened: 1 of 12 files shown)


tests/conftest.py
```diff
@@ -8,9 +8,26 @@
...
+       a) we don't stomp on existing session directories
+       b) we don't require run time configuration
+       c) we ensure that everything gets properly cleaned up after the test run
+    '''
+    dirname = py.path.local(mkdtemp())
+    request.addfinalizer(lambda: dirname.remove(rec=1, ignore_errors=True))
+    return str(dirname)
+
+@pytest.fixture(scope="module")
+def static_url_space():
```
basilveerman: This is not used and frontend is not defined in this module. What was the plan for this?

tests/conftest.py
```diff
@@ -8,9 +8,26 @@
...
+    return frontend
```
basilveerman: static_url_space does not appear to be used and frontend is not defined in this module. What was the plan for this?

jameshiebert:
Yep, you're right. It was used earlier in the branch, but was abandoned. Fixed with 96b9d6e902cd2332b717f2b75b4ca4c072d4ea4d

basilveerman:
These changes look great. I don't see there being issues since it is really only the instantiation method that has changed rather than any internal structure of the application. The exported WSGI containers are really the same. I like that you've changed to source config.yaml from the PDP_CONFIG environment variable and would say it's actually *less* deployment steps since it removes the need to edit that on every deploy.

I'll go through the full install/test procedure and then merge in for 2.3.0rc2

PR #30, opened 2015-06-17
Feature/2.3.x- oauth

basilveerman:


pdp/auth.py
```diff
@@ -0,0 +1,61 @@
...
+    if request.method == 'POST':
+        email = request.POST.get('email', None)
+        if not email:
+            start_response('400 Bad Request', [('Content-type', 'text/html; charset=utf-8')])
+            return ['Email must be included in POST body']
+        session['email'] = email
+        start_response('200 OK', [('Content-type', 'text/html; charset=utf-8')])
+        return json.dumps({'session_id': session.id})
+
+    if request.method == 'GET':
```
jameshiebert: There are other available HTTP verbs that this method doesn't support, so probably better to just say

```python
else:
    start_response('400 BAD REQUEST', [('Content-type', 'text/html; charset=utf-8')])
    return ['Login handler does not support {} requests'.format(request.method)]
```

or `"Login handler only supports POST requests"`

pdp/auth.py
```diff
@@ -0,0 +1,61 @@
...
+    '''
+
+    session = environ.get('beaker.session', {})
+    email = session.get('email', None) 
+    if email:
+        session.delete()
+        start_response('200 OK', [('Content-type', 'text/html; charset=utf-8')])
+        return ['Sucessfully logged out']
+
+def user_manager():
```
jameshiebert: This is a good/clean way to organize the URL space. I like it.

pdp/config.yaml
```diff
@@ -1,17 +1,15 @@
...
-geoserver_url: http://atlas.pcic.uvic.ca/geoserver/ # change for production
+geoserver_url: https://atlas.pcic.uvic.ca/geoserver/ # change for production
 ncwms_url: 
-    - http://atlas.pcic.uvic.ca/ncWMS/wms # change for production
+    - https://atlas.pcic.uvic.ca/ncWMS/wms # change for production
 tilecache_url: 
-    - http://a.tiles.pacificclimate.org/tilecache/tilecache.py
-    - http://b.tiles.pacificclimate.org/tilecache/tilecache.py
-    - http://c.tiles.pacificclimate.org/tilecache/tilecache.py
+    - http://tiles.pacificclimate.org/tilecache/tilecache.py
```
jameshiebert: Any reason in particular for not splitting out the tile servers anymore?

pdp/main.py
```diff
@@ -56,6 +57,7 @@ def initialize_frontend(global_config, use_auth=False, use_analytics=False):
     wsgi_app = PathDispatcher([
         ('^/css/(default|pcic).css$', static.Cling(resource_filename('pdp_util', 'data'))), # a bit of a hack for now
         ('^/check_auth_app/?$', check_auth),
```
jameshiebert: With /user/profile is the `check_auth_app` now obsolete/unnecessary? If so, it should be removed from the code and cleaned up.

pdp/static/js/pdp_auth.js
```diff
@@ -41,117 +24,132 @@ window.pdp = (function (my, $) {
...
+                "email": email
             }
+        })
+        // FIXME Add failure handler
+    }
+
+    my.logout = function(network) {
+
+        // oauth logout
+        if (network) {
```
jameshiebert: Soooo, can one call `logout()` with no network to force a request to /user/logout and show the loginButton? Or what happens if `logout()` is called with no network defined?

pdp/static/js/pdp_auth.js
```diff
@@ -185,79 +183,133 @@ window.pdp = (function (my, $) {
         pdp.createCookie(name, "", -1);
     };
 
-    my.getLoginForm = function (providers) {
+    my.validateEmail = function (email) {
+        var re = /^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,6}(?:\.[a-z]{2})?)$/i;
```
jameshiebert: Since you're not capturing any groups can you remove group parens that aren't modified by number parameters? Make it a little easier to read? For example in the user section this:
```([\w-]+(?:\.[\w-]+)*)```
can be replaced with this:
```[\w-]+(?:\.[\w-]+)*``` which is the same if you're not capturing groups.

File: pdp/static/js/pdp_auth.js [hunk 7]
```diff
@@ -269,4 +321,21 @@ window.pdp = (function (my, $) {
     };
 
     return my;
-}(window.pdp, jQuery));
+}(window.pdp, jQuery, hello));
+
+var GOOGLE_CLIENT_ID = {
```
jameshiebert: Do these need to be protected at all?

tests/conftest.py
```diff
@@ -55,29 +55,12 @@ def check_auth_app(session_dir):
     return SessionMiddleware(check_auth, auto=1, data_dir=session_dir)
 
 @pytest.fixture(scope="module")
-def authorized_session_id(check_auth_app):
-    # FIXME: I shouldn't have to do this, but the store doesn't get initialized until the first request
+def authorized_session_id(pcic_data_portal):
```
jameshiebert: This is soooo much simpler and less prone to breakage. Thank goodness. The old way was a mess; good work.

requirements.txt
```diff
@@ -18,7 +18,7 @@ itsdangerous==0.24
 modelmeta==0.0.5
 numpy  # Allow pip or system install for travis
 openid2rp==1.12
-pdp-util==0.2.4
+pdp-util==0.2.5-dev
```
jameshiebert: We need a version of `pdp-util` that actually satisfies this in our pypi server.

jameshiebert:
Since we still don't yet have travis running on the full test suite, could you attach a log of running the full test suite on this branch? I get somewhere in the neighborhood of 33 pass / 16 fails, from this:

pdp/config.yaml
basilveerman: Yes: SSL wildcard certs. Can't have *.*.rootdomain.org, only *.rootdomain.org. Should actually be https, but the tileserver wasn't live over https at the time.

pdp/static/js/pdp_auth.js
basilveerman: Public key is required in the authorization request. The Oauth provider is configured with a set redirect uri (per app/host) that will only send the auth token back to the configured endpoint. We do not use explicit grant/auth code grant which would require secret keys (projected) and server-server authorization.

PR #33, opened 2015-08-13
Merge Hotfix/2.3.4 evented ncwms capabilies. Fixes #32.

### LINKED ISSUE #32: Race condition for climate layer capabilities document when changing layers on raster portals
2.3.3 made changes to raster portals to trigger the download link to update the selection area on any climate layer change. This was not successful due to event handlers different bound object expectations (dlLink.onBoxchange vs dlLink.onLayerChage) and revealed a race condition for the climate layer Capabilities document.

Layer change logic needs to deal with these issues and remove the possibility of race conditions inherent to using global variables.

basilveerman:
Overview

* ncWMS layer capabilities now refreshed on layer change event
* `catalog` removed from global scope
* removed various unused global vars: `current_dataset`, `ensemble_name`, `gs_url`, `tilecache_url`, `map`, `pcds_map`
* Unified `catalog` request method between portals
* Unified `ncwms` declaration in global scope

### DIFF
--- a/pdp/static/js/bccaq_extremes_app.js
+++ b/pdp/static/js/bccaq_extremes_app.js
@@ -34,9 +34,18 @@ $(document).ready(function () {
             dlLink.onExtensionChange($(this).val());
         }
     );
-    ncwmsLayer.events.register('change', dlLink, dlLink.onLayerChange);
-    ncwmsLayer.events.register('change', dlLink, dlLink.onBoxChange);
+
+    ncwmsLayer.events.register('change', dlLink, function () {
+        getNCWMSLayerCapabilities(ncwmsLayer).done(function() {
+            if (selectionLayer.features.length > 0) {
+                dlLink.onBoxChange({feature: selectionLayer.features[0]});
+            }
+        });
+    });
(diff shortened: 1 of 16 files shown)


jameshiebert:
So I have two minor comments and then a blocker comment. (I'll push patches for the minor comments)

## Minor

* I see that you moved the `var` statement down to where ncwms is first assigned. jslint tells us not to do that, and to keep all of the variable declarations at the top of the function in a single `var` statement.
* I know that I used `registerPriority()` in my first stab at this problem, but really, `registerPriority()` is just a hack that (inadequately) compensates for the fact that we had a race condition. If we have fixed the race condition, we shouldn't require any prioritization in the event queue and it should be truly asynchronous.

## Blocker

When the user first draws a selection box, with no layer change taking place, the onBoxChange event does not complete. It errors out at pdp_controls.js:286:
```js
raster_proj = getRasterNativeProj(ncwmsCapabilities, lyr_id);
```
with `ncwmsCapabilities` not being defined. It might be sufficient, to just need to make that same promise at the top of every *app.js page?

Issue #43, opened 2016-09-23
Colorscale Legend


tqmurdock:
Numbers on legend should be rounded to nearest mm and nearest degree Celsius

jameshiebert:
To be more specific, this affects the colorscale bar legend on the raster pages within `pdp`. I.e. the Colorbar object [here](https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/pdp_controls.js#L115).

PR #44, opened 2016-10-06
Set colorbar precision according to range

### LINKED ISSUE #41: Documentation for color scalebar does not exist
The color scalebar on the raster porter was added in release 2.2.0, but no documentation was added to the user docs. It shouldn't _need_ documentation, but it does have some non-obvious bevhaviours.
1. The displayed range is for the entire dataset (e.g. all of time)
2. Range is not configurable.
3. The scale is linear for temperature variables, but logrithmic for precipitation.

These things (and possibly others) need to be documented for the users.

CBMasri:
Fixes #41, #43

Values are calculated based on the range with the following behaviour (R = range):

- R > 10,000               --> Round to nearest hundred
- 100 < R <= 10,000   --> Round to nearest 10
- 10 < R <= 100          --> Round to nearest integer
- 1 < R <= 10              --> Round to 2 significant digits
- R < 1                        --> Round to 3 significant digits


jameshiebert:
Hey Carl, can you build a docker image tagged "pdp-pr-44" and deploy it to `docker1`?

pdp/static/js/pdp_controls.js
```diff
@@ -179,6 +180,35 @@ Colorbar.prototype = {
...
+    },
+
+    apply_precision: function (sig) {
+      this.minimum = this.minimum.toFixed(sig);
+      this.maximum = this.maximum.toFixed(sig);
+      this.midpoint = this.midpoint.toFixed(sig);
+    },
+
+    estimate_precision: function () {
+      var range = this.maximum - this.minimum;
```
jameshiebert: Could you add some comments here to describe the intent of the return value? You're kind of doing two things here: mutating state and returning a value. It would be better to perform one (and only one) operation. Also, why not pass in min/max as parameters, instead of retrieving from properties? Try and use as little state as possible. In this case you don't need any.

pdp/static/js/pdp_controls.js
```diff
@@ -117,6 +117,7 @@ function Colorbar(div_id, layer) {
     this.layer = layer;
     this.minimum = 0.0;
     this.maximum = 0.0;
+    this.midpoint = 0.0;
```
jameshiebert: So it may have seemed odd to you that min/max are saved as state while the midpoint is not. But the reason that getting the min/max is a relatively expensive operation (an async network query), while the midpoint is simple client-side calculation. I would prefer to keep it that way. Maintain as little state as possible. At the very least, midpoint should be a variable local to the function that is using it (`redraw()`).

pdp/static/js/pdp_controls.js
```diff
@@ -131,7 +132,7 @@ function Colorbar(div_id, layer) {
 Colorbar.prototype = {
     constructor: Colorbar,
 
-    midpoint: function () {
+    calculate_midpoint: function () {
```
jameshiebert: I think this is appropriate to rename this function, as a way to clarify that it's actually doing something and it's a function, not a property.

pdp/static/js/pdp_controls.js
```diff
@@ -179,6 +180,35 @@ Colorbar.prototype = {
         }
     },
 
+    round_to_nearest: function (mult) {
+      this.minimum = Math.round(this.minimum / mult) * mult;
```
jameshiebert: Again, I think that this function should be structured as a take_parameter/return_value, rather than a side-effects based function that modifies state. Since you're performing the exact operation for each value, why not just pass the value in as a parameter and return the result? E.g.

```javascript
round_to_nearest: function (value, mult) {
    return Math.round(value / mult) * mult;
},
```

pdp/static/js/pdp_controls.js
```diff
@@ -204,15 +234,16 @@ Colorbar.prototype = {
...
 
     redraw: function () {
+        this.midpoint = this.calculate_midpoint();
+        this.apply_precision(this.estimate_precision());
         var div = $("#" + this.div_id);
         div.css('background-image', "url(" + this.graphic_url() + ")");
-        div.find("#minimum").html(round(this.minimum) + " " + this.units);
-        div.find("#maximum").html(round(this.maximum) + " " + this.units);
-        div.find("#midpoint").html(round(this.midpoint()) + " " + this.units);
+        div.find("#minimum").html(this.minimum + " " + this.units);
```
jameshiebert: In these three lines, I'd rather you apply the precision locally at render-time than modify it in place and then render the modified version.

pdp/static/js/pdp_controls.js
CBMasri: Sure thing.

pdp/static/js/pdp_controls.js
CBMasri: Agreed, I'll split this into two functions.

pdp/static/js/pdp_controls.js
CBMasri: Sure, no problem.

pdp/static/js/pdp_controls.js
jameshiebert: Yes, but functions are meant to be called multiple times :) That's part of what tips you off to the fact that something should be a function... it's reused over and over.

CBMasri:
Docker image pdp-pr-44 has been created and a container of the same name is running on docker1 (http://docker1.pcic.uvic.ca:8080/bc_prism/map/). Hop into it with `docker attach pdp-pr-44`.

jameshiebert:
> Docker image pdp-pr-44 has been created

Great, thanks! Since this change affects (to a small degree) the way that the team leads' data are presented, I'm going to send a quick note to them, asking that they play with the demo and determine whether the changes work for them. Could you add a description to this PR that explains the method by which you round the values, so everyone can provide feedback on whether that's acceptable for their data? Then we can tweak the calculations if necessary.

PR #45, opened 2016-10-07
Added colorbar legend documentation to user docs

CBMasri:


jameshiebert:
Thanks for adding this documentation (it looks great), but this should probably be included in #44. If you rebase this change onto the feature/colorbar-precision branch and push, you can include it in that PR.

CBMasri:
Sure, no problem.

PR #46, opened 2016-10-12
Feature/colorbar sync

### LINKED ISSUE #42: Front-end colorscale bar is out of sync with WMS requests
Francis has noted that on the [downscaled GCMs data portal page](http://tools.pacificclimate.org/dataporta/downscaled_gcms/map/), the color scale range goes from -61 to +47 degrees celsius, but it's displaying a map for January 2000 and showing values near the top end of the range.

screen shot 2016-09-21 at 10 07 04 am

Temperature > 20 is implausible for Vancouver Island in January, so something is up with the map.

It turns out that if you look at the WMS request headers:
http://tools.pacificclimate.org/ncWMS-PCIC/wms?LAYERS=pr-tasmax-tasmin_day_BCSD-ANUSPLIN300-CanESM2_historical-rcp26_r1i1p1_19500101-21001231/tasmax&TRANSPARENT=true&STYLES=boxfill/ferret&TIME=2000-01-01&NUMCOLORBANDS=254&VERSION=1.1.1&SRS=EPSG:4326&COLORSCALERANGE=-50,11&LOGSCALE=false&SERVICE=WMS&REQUEST=GetMap&FORMAT=image/png&BBOX=-129.375,52.03125,-127.96875,53.4375&WIDTH=512&HEIGHT=512
the COLORSCALE range is fixed at something different (-50 to 11) than the values of the scale bar.

If I recall correctly, the COLORSCALE range is hardcoded ([here](https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/canada_ex_map.js#L40)) because people wanted to see _some_ spatial variation on the map (rather than fixing the range to be the full range of the dataset and then having a small range show up for a single time step). But this is a little schizophrenic: we need to use the same approach (fully dynamic, or fully hard-codeded) for both the WMS params _and_ the colorscale bar. Otherwise, they'll continue to be out of sync.

Action on this ticket requires deciding on the approach and then implementing it.

CBMasri:
Fixes #42

### DIFF
--- a/pdp/static/js/canada_ex_map.js
+++ b/pdp/static/js/canada_ex_map.js
@@ -32,12 +32,10 @@ function init_raster_map() {
         layers: defaults.dataset + "/" + defaults.variable,
         transparent: "true",
         styles: "boxfill/ferret",
-        // colorscalerange: 'auto', //FIXME: after layer update, set colorscalerange based on map extent
         time: "2000-01-01",
         numcolorbands: 254,
         version: "1.1.1",
         srs: "EPSG:4326",
-        colorscalerange: "-50,11.0",
         logscale: false
     };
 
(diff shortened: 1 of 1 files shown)


CBMasri:
No problem: http://docker2.pcic.uvic.ca:8080/downscaled_gcms/map/

PR #47, opened 2016-10-14
Feature/dockerize

CBMasri:
Adds the files to run dockerized pdp and nginx reverse-proxy.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,7 +21,7 @@ ENV CPLUS_INCLUDE_PATH /usr/include/gdal
 ENV C_INCLUDE_PATH /usr/include/gdal
 ENV PDP_CONFIG /root/pdp_config.yaml
 ENV PATH /root/pdp/pyenv/bin:$PATH
-ARG BRANCH=master
+ARG BRANCH=dev
 
 # Clone repo and set up virtualenv
 WORKDIR /root/
@@ -40,7 +40,6 @@ RUN sed -i 's/netCDF4==1.1.1/netCDF4/' test_requirements.txt
 # Install dependencies
 RUN pip install -i https://pypi.pacificclimate.org/simple/ \
     -r requirements.txt \
-    -r data_format_requirements.txt \
     -r test_requirements.txt \
     sphinx \
     gunicorn \
@@ -65,5 +64,5 @@ EXPOSE 8000 8001
 # Build template files
 ENTRYPOINT ["/root/pdp/docker-entrypoint.sh"]
 
-# CMD ["supervisord", "-c", "/etc/supervisord.conf"]
-CMD ["/bin/bash"]
+CMD ["supervisord", "-c", "/etc/supervisord.conf"]
+# CMD ["/bin/bash"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,69 @@
+############################################
+# Dockerfile to run the PCIC data portal   #
+# Based on Ubuntu 16.04                    #
+############################################
+
+FROM ubuntu:16.04
+MAINTAINER Carl Masri <cmasri@uvic.ca>
+
+# Get the system level dependencies
+RUN apt-get update && apt-get install -y \
+    python-dev \
+    python-virtualenv \
+    build-essential \
+    libhdf5-dev \
+    libgdal-dev \
+    libnetcdf-dev \
+    git
+
+# Set up environment variables
+ENV CPLUS_INCLUDE_PATH /usr/include/gdal
+ENV C_INCLUDE_PATH /usr/include/gdal
+ENV PDP_CONFIG /root/pdp_config.yaml
+ENV PATH /root/pdp/pyenv/bin:$PATH
+ARG BRANCH=master
+
+# Clone repo and set up virtualenv
+WORKDIR /root/
+RUN git clone https://github.com/pacificclimate/pdp -b ${BRANCH}
+WORKDIR /root/pdp/
+RUN virtualenv pyenv
+
+# Install local dependencies and fix SSL issue
+RUN pip install numpy requests[security]
+# Separate RUN statement for gdal install is required!
+RUN pip install gdal==1.11.2
+
+# Remove netCDF4 version requirement
+RUN sed -i 's/netCDF4==1.1.1/netCDF4/' test_requirements.txt
+
+# Install dependencies
+RUN pip install -i https://pypi.pacificclimate.org/simple/ \
+    -r requirements.txt \
+    -r data_format_requirements.txt \
+    -r test_requirements.txt \
+    sphinx \
+    gunicorn \
+    gevent \
+    supervisor \
+    j2cli[yaml]
+
+# Install and build the docs
+RUN python setup.py install
+RUN python setup.py build_sphinx
+RUN python setup.py install
+
+# Create directory for supervisord logs
+RUN mkdir etc/
+
+# Add the template config files
+COPY templates/ /templates/
+COPY docker-entrypoint.sh /root/pdp/
+
+EXPOSE 8000 8001
+
+# Build template files
+ENTRYPOINT ["/root/pdp/docker-entrypoint.sh"]
+
+# CMD ["supervisord", "-c", "/etc/supervisord.conf"]
+CMD ["/bin/bash"]
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -1,0 +1,16 @@
+version: '2'
+services:
+  pdp:
+    image: pdp
+    container_name: pdp
+    ports:
+      - "8000:8000"  # frontend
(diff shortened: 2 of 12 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+# Set up environment variables
+ENV CPLUS_INCLUDE_PATH /usr/include/gdal
+ENV C_INCLUDE_PATH /usr/include/gdal
+ENV PDP_CONFIG /root/pdp_config.yaml
+ENV PATH /root/pdp/pyenv/bin:$PATH
+ARG BRANCH=dev
+
+# Clone repo and set up virtualenv
+WORKDIR /root/
+RUN git clone https://github.com/pacificclimate/pdp -b ${BRANCH}
```
jameshiebert: So, I don't think that this is what we discussed over e-mail. We want docker to build whatever branch is currently checked out. The easiest way to do this is to just ADD the source dir to the image, like we do [here](https://github.com/pacificclimate/climate-explorer-backend/blob/master/Dockerfile#L3).

This will also mean that you don't need to install git in the system depends.

docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+# Based on Ubuntu 16.04                    #
+############################################
+
+FROM ubuntu:16.04
+MAINTAINER Carl Masri <cmasri@uvic.ca>
+
+# Get the system level dependencies
+RUN apt-get update && apt-get install -y \
+    python-dev \
+    python-virtualenv \
```
jameshiebert: Actually, after further consideration, virtualenv is not really necessary, since we're doing this all in a container, right? May as well just install python-pip and do "system" install inside the container with pip.

docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+    build-essential \
+    libhdf5-dev \
+    libgdal-dev \
+    libnetcdf-dev \
+    git
```
jameshiebert: Don't need git here.

jameshiebert: I think we also need to apt-get install libffi-dev, for requests[security] to install properly. (At least I get lots of warnings w/o it).

docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+RUN pip install gdal==1.11.2
+
+# Remove netCDF4 version requirement
+RUN sed -i 's/netCDF4==1.1.1/netCDF4/' test_requirements.txt
+
+# Install dependencies
+RUN pip install -i https://pypi.pacificclimate.org/simple/ \
+    -r requirements.txt \
+    -r test_requirements.txt \
+    sphinx \
```
jameshiebert: Perhaps all that these extras that you have added could be put into a "deploy_requirements.txt" file, just so that it's clear why they are needed.

docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+    gunicorn \
+    gevent \
+    supervisor \
+    j2cli[yaml]
+
+# Install and build the docs
+RUN python setup.py install
```
jameshiebert: I don't think that you need this first 'install', if you do it again after you build the docs.

docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
+############################################
+# Dockerfile to run the PCIC data portal   #
+# Based on Ubuntu 16.04                    #
+############################################
+
+FROM ubuntu:16.04
+MAINTAINER Carl Masri <cmasri@uvic.ca>
+
+# Get the system level dependencies
+RUN apt-get update && apt-get install -y \
```
jameshiebert: It might be good to separate out the "apt-get update" and the install into two separate RUN commands. That way we get the cached image after an apt-get update, and during development, any time that we need to change the system depends, we don't have to run the apt-get update.

This advice can be generalized... in your Dockerfile always do the more expensive operations as early as possible to get the best caching benefits. Obviously order is important, but in places where it doesn't matter, use cost as a tie breaker as for which operation goes first.

docker/proxy/Dockerfile
```diff
@@ -0,0 +1,17 @@
+###############################################
```
jameshiebert: I actually noticed that nginx has [their own official Docker image](https://hub.docker.com/r/_/nginx/) that should be able to do everything that you're doing here. Any reason not to use it instead?

docker/docker-compose.yaml
```diff
@@ -0,0 +1,16 @@
+version: '2'
```
jameshiebert: I got errors trying to run docker compose on this file and it turns out that ubuntu has an older (pre 1.6) version of docker compose that doesn't support [versioning](https://docs.docker.com/compose/compose-file/#/versioning). It might be useful to either provide a version 1 docker compose file, or put in the deployment guide that this requires [docker-compose > 1.6](https://docs.docker.com/compose/compose-file/#/version-2).

docker/Dockerfile
jameshiebert: Ah good call, Basil. That `get_distribution` call actually kind of bugs me (even though I wrote it), both in the docs and in the web page templates. B/c let's say you're reviewing changes in a development branch, or you've deployed a demo of something that you want to show others before releasing it... the stated version is kind of a lie. Really what you want is a commit hash, tag, or branch name. Any good ideas on how to make that work?

basilveerman: Nothing good. I ended up _removing_ a different method I tried in cfmeta due to pypi/readthedocs not liking using a VERSION.txt file: https://github.com/pacificclimate/cfmeta/commit/741bf1c8ef04046ead4e4151a9ab4f4c10eccf37.

There may have been a better way around that however.

docker/Dockerfile
CBMasri: Good call.

docker/Dockerfile
CBMasri: Yes I was noticing that as well. I'm actually going to see if I can remove the requests[security], or at least just get the minimal number of required packages to fix the pypi SSL issue.

docker/Dockerfile
CBMasri: Sure, that's a good idea.

docker/Dockerfile
CBMasri: Are you sure? According to the [README](https://github.com/pacificclimate/pdp/commit/9bfdf76516ac230081876474f492588cca118c4c) it's needed, although I haven't tried removing it (but I will).

docker/proxy/Dockerfile
CBMasri: I found it was useful to be able to hop into the container and edit the configuration files and stop/start nginx, which isn't possible from the official image. Now that the configuration has been sorted I can update the Dockerfile with the official image.

docker/docker-compose.yaml
CBMasri: Okay, good to know. I'll add the version req into the deployment guide.

docker/Dockerfile
jameshiebert: OK, I have a good idea for how to solve this programmatically, but I think that it's beyond the scope of this PR. I'll start a new issue for it... for now @CBMasri feel free to leave in the first install line. Thanks for pointing out this issue @basilveerman

CBMasri:
Updated dockerfiles have been pushed, let me know what you think!
Also, I've rebased ``contributing-update``, did you want me to merge those changes into ``feature/dockerize`` so that they show up in this PR?

jameshiebert:
> did you want me to merge those changes into `feature/dockerize` so that they show up in this PR?

Yes please! :)

docker/proxy/nginx.conf
```diff
@@ -0,0 +1,35 @@
...
+}
+
+http {
+
+  server {
+    listen 8080;
+    server_name "";
+
+    location / {
+      proxy_pass					http://docker2.pcic.uvic.ca:8000;
```
jameshiebert: Can you audit the config files and ensure that the docker host hostname (e.g. docker2.pcic) is templated out and not hard-coded? You can either use your existing j2 templating scheme or see the section "Using environment variables in nginx configuration" in the ["nginx docs"](https://hub.docker.com/r/_/nginx/).

CBMasri: Ah yes, this is definitely a good idea. For simplicity, I think nginx's solution will work just fine.

CBMasri:
Okay, containers derived from the Nginx image can now be run dynamically using env variables. I wasn't crazy on how their method *required* docker-compose, so I changed it slightly so that the container can be started directly using ``docker run ...``. There isn't a way to provide default values as a fallback (without adding them directly into the Dockerfile) which is a little unfortunate, but the ``docker-compose`` file at least provides a bit of a short cut rather than having to specify the variables each time.

jameshiebert:
OK, I added a few final changes that I think shore this up. Please look over them, so that you are aware of the implications. We still need a few more details in the contributors guide, but I think that those can be added at a later time.

PR #49, opened 2016-10-19
Add info about environment config

CBMasri:


CONTRIBUTING.rst
```diff
@@ -137,16 +212,16 @@ By default, the pdp Dockerfile exposes port 8000 (the port that gunicorn will ru
...
 
 Data Volume Container
 ^^^^^^^^^^^^^^^^^^^^^
 
-Not all data is accessible to the pdp remotely, some of it (the hydro station output, for example) is stored in the host environment. Docker provides a nice utility called ``volumes`` which makes host directories accessible to Docker containers, but to avoid constantly having to specify the paths when creating a new Docker container we can use what's called a "data volume container". The following command will create a data volume container and mount the target host directories (most likely in /storage/data/). ``:ro`` signifies that this is a "read-only" volume.
+Not all data is accessible to the pdp remotely, some of it (the hydro station output, for example) is stored in the host environment. Docker provides a nice utility called ``volumes`` which makes host directories accessible to Docker containers, but to avoid constantly having to specify the paths when creating a new Docker container we can use what's called a "data volume container". Target host directories are mounted using the ``-v`` option. Docker volumes default to mount in read-write mode, but this can be changed to read-only by appending ``:ro``.
```
jameshiebert: I do like the way that you stated your change: "mounted using the -v option". But I think that the docs should indicate that data volumes should only be mounted "ro". There is no reason for a data publishing app to mount the volumes "rw" and doing so would be a security hazard, so it's probably appropriate for the docs to emphasize that point.

CONTRIBUTING.rst
```diff
@@ -137,16 +212,16 @@ By default, the pdp Dockerfile exposes port 8000 (the port that gunicorn will ru
...
 
-.. FIXME The data volume containers stop/exit immediately. Describe how to prevent this.
+The following command will create a data volume container. This should only need to be run once, as data volumes in docker are persistent and will remain even after the container has exited.
```
jameshiebert: Interesting; I did not know this. Thanks for clarifying!

PR #52, opened 2016-11-02
Feature/show dev branch

### LINKED ISSUE #48: Exact development branch/commit should be visible
In several places (e.g. in the [documentation](https://github.com/pacificclimate/pdp/blob/master/doc/source/conf.py#L55) and on the [web page templates](https://github.com/pacificclimate/pdp/blob/master/pdp/templates/map.html#L65)), we show a package "version number". However, in most cases this version number is only the latest child release number, since it is created from [manually updating a release number in `setup.py`](https://github.com/pacificclimate/pdp/blob/master/setup.py#L33). In most cases, the state version is kind of a lie.

There are many instances in which one would want to see the _exact_ commit hash or branch. For example:
- Reviewing changes in a development branch,
- Reviewing a deployed demo of something new feature that you want to show others before releasing it
- Reviewing documentation before releasing
- Running parallel versions of the software (e.g. a stable release and a dev release)

It would be best if these could be generated programmatically. Ideas?

``` bash
$ git rev-parse --short HEAD
e140dc4
$ git symbolic-ref HEAD | sed -e 's/refs\/heads\///'
feature/dockerize
```

I'm not sure when this should actually happen... if it's done at package run time, this doesn't make sense because it's not running from a git repo. Maybe best to use [`GitPython`](https://pypi.python.org/pypi/GitPython/python-git) and put this in the setup.py script somewhere?

CBMasri:
Fixes #48. Shows the branch and commit on the dataportal

pdp-commit

as well as in the docs

pdp-docs


Dockerfile
```diff
@@ -6,14 +6,14 @@
 FROM ubuntu:16.04
 MAINTAINER Carl Masri <cmasri@uvic.ca>
 
-RUN apt-get update
-RUN apt-get install -y \
+RUN apt-get update && apt-get install -y \
```
CBMasri: This was changed to comply with Dockerfile [best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get).

doc/source/conf.py
```diff
@@ -51,10 +51,11 @@
...
 #
 from pkg_resources import resource_filename, get_distribution
+from pdp import sha, branch
 # The full version, including alpha/beta/rc tags.
-release = get_distribution('pdp').version
-# The short X.Y version.
-version = release.rsplit('.', 1)[0]
+version = get_distribution('pdp').version
+# The exact release (branch:sha)
+release = "%s (%s)" % (branch, sha)
```
CBMasri: These may need to be swapped, I wasn't sure about the proper semantics.

pdp/__init__.py
```diff
@@ -15,6 +15,7 @@
 import yaml
 import static
 from beaker.middleware import SessionMiddleware
+from git import Repo
```
jameshiebert: So, I initially liked this approach. However, the big, big drawback is that with this change, you would *always* need to run the code out of an existing git repository. This would basically eliminate the ability to run the code as a [source distribution](https://packaging.python.org/distributing/#source-distributions) or a built [wheel](https://packaging.python.org/distributing/#wheels).

I think that it would be a better choice to put all of the `get_commitish()` code in the setup.py to be run at build time or doc build time. That way we have the option of building our package on a dev machine, pushing the build to our PyPI repo, and installing from there on a prod machine (that wouldn't necessarily have git installed).

pdp/__init__.py
```diff
@@ -56,7 +68,8 @@ def get_config():
             'js/pdp_vector_map.js'
             ], debug=(not config['js_min'])),
         'templates': resource_filename('pdp', 'templates'),
-        'version': get_distribution('pdp').version
+        'version': get_distribution('pdp').version,
+        'revision': "%s:%s" % (branch, sha)
```
jameshiebert: I'm not sure if this can be baked in to the package metadata at build time, but a [local version identifier](https://packaging.python.org/distributing/#local-version-identifiers) might be your best bet...

pdp/__init__.py
CBMasri: Good points! My first attempt _did_ put the `get_commitish()` code into setup.py, however, this caused issues when trying to collect the values with `get_distribution()` as the added keyword (`revision`) is non-standard and did not get added to the package metadata. I'll take a look at the local version identifier and see if that will work instead.

pdp/__init__.py
```diff
@@ -56,14 +57,26 @@ def get_config():
...
         }
 
     config = updateConfig(global_config, config)
     if config['session_dir'] == 'default':
         config['session_dir'] = resource_filename('pdp', 'pdp_session_dir')
     return config
 
+def parse_version(type):
+    full_version = get_distribution('pdp').version
+    regex = ur"^((?:\w+\.?)+)\+?(.*)\.(\w{6})$"
```
jameshiebert: It's usually good to write some tests against any regex that you write.

pdp/__init__.py
```diff
@@ -56,14 +57,26 @@ def get_config():
...
+        'revision': parse_version("revision")
         }
 
     config = updateConfig(global_config, config)
     if config['session_dir'] == 'default':
         config['session_dir'] = resource_filename('pdp', 'pdp_session_dir')
     return config
 
+def parse_version(type):
+    full_version = get_distribution('pdp').version
```
jameshiebert: If you split out the get_distribution call to be outside of this function, it makes it easier to test (b/c you can pass in an arbitrary version as a parameter in the tests).

pdp/__init__.py
```diff
@@ -56,14 +57,26 @@ def get_config():
...
+    regex = ur"^((?:\w+\.?)+)\+?(.*)\.(\w{6})$"
+    matches = re.match(regex, full_version)
+    if matches:
+        if type == "version":
+            return matches.group(1)
+        else if type == "revision":
```
jameshiebert: Use an `elif` here.

setup.py
```diff
@@ -29,15 +30,22 @@ def find():
                 yield os.path.join(dirname, filename).lstrip(pkg_dir)
     return [ x for x in find() ]
 
+def get_commitish():
+    repo = Repo(os.getcwd())
+    sha = repo.head.object.hexsha
+    branch = repo.active_branch.name
+    safe_branch = branch.replace('/', '.').replace('-', '.').replace('_', '.')
+    return "%s.%s" % (safe_branch, repo.git.rev_parse(sha, short=6))
```
jameshiebert: % formatting isn't used as much anymore. Better to use string.format().

jameshiebert:
I already merged this, but turns out that the [build fails](https://travis-ci.org/pacificclimate/pdp/builds/173398106) on TravisCI... looking into why...

PR #54, opened 2016-11-04
Add IOError handling while attempting to read the config file

### LINKED ISSUE #53: Improve config process and error message
At present, if a developer fails to provide a config file for pdp, you get this obnoxious error message:

```python
../../../virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pdp/__init__.py:77: in wrap_auth

    config = get_config()

../../../virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pdp/__init__.py:28: in get_config

    with open(config_filename) as f:

E   IOError: [Errno 2] No such file or directory: '/var/www/dataportal/config.yaml'
```

That stops you in your tracks. I think that we should consider either providing an in-package resource file for the default config, and just give a warning if it falls back to using it. Or at the very least, we should catch this `IOError` and give the user instructions on the fact that they need to provide a config file.

CBMasri:
Fixes #53.

### DIFF
--- a/pdp/__init__.py
+++ b/pdp/__init__.py
@@ -5,7 +5,7 @@
 
 __all__ = ['get_config', 'wrap_auth']
 
-import os
+import sys, os
 from os.path import dirname
 import atexit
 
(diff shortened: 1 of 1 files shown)


jameshiebert:
Looks good to me

PR #56, opened 2016-11-07
Feature/raster single cell download

### LINKED ISSUE #8: Write a control which generates a sinlge-cell OPeNDAP link on map click
Affected pages: all raster pages

At present the entire design of the system and the nature of how large high-resolution spatio-temporal data can be it is inconvenient to get a timeseries of data for a single grid cell.

That said, it is _possible_, albeit time consuming. We should add a control on the web UI which takes the coordinates of a mouse click and uses them to generate a single-cell OPeNDAP link, for example:

http://tools.pacificclimate.org/dataportal/downscaled_gcms/data/pr+tasmax+tasmin_day_BCSD+ANUSPLIN300+CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc.nc?tasmax[0:55152][89:89][212:212]&

For pages where we enable timeseries plotting via the mouse click ([bc_prism](http://tools.pacificclimate.org/dataportal/bc_prism/map/) and ([downscaled_gcm_extremes](http://tools.pacificclimate.org/dataportal/downscaled_gcm_extremes/map/), we could add this link to the plot window.

CBMasri:
Fixes #8. This PR introduces a new point tool to the control panel of the raster portals which allows for downloading a single-cell OPeNDAP link. When selected, the user can click anywhere on the raster overlay to generate a prompt displaying the precise download link.

dl-prompt

This control will adhere to any other filters applied (eg. a certain date range).

[Demo](http://docker2.pcic.uvic.ca:8080/bc_prism/map/)

*Currently does not support ARC/Info ASCII Grid output format.*


pdp/static/js/pdp_controls.js
```diff
@@ -290,6 +297,36 @@ RasterDownloadLink.prototype = {
         );
         return url;
     },
+    getSingleCellDownloadLink: function (bounds, feature) {
+        if (this.ext !== "aig") {
```
jameshiebert: Since the aig format is the exceptional case, it's sometimes a little cleaner to make a *positive* test for the aig format, and then return. Then the bulk of the method's code isn't stuck in an `if` branch.

pdp/static/js/prism_demo_map.js
```diff
@@ -99,7 +98,7 @@ function init_prism_map() {
         [
             ncwms,
             selectionLayer,
-            getBC3005OsmBaseLayer(pdp.tilecache_url, 'BC OpenStreeMap', 'bc_osm')
+            getBC3005OsmBaseLayer(pdp.tilecache_url, 'BC OpenStreetMap', 'bc_osm')
```
jameshiebert: Good catch! Thanks!

doc/source/raster.rst
```diff
@@ -28,7 +28,19 @@ To select a spatial subset for data download, one must first switch to the polyg
 
 .. figure:: images/draw_polygon_on.png
 
-When in spatial subset selection mode, you can select a rectangular area by doing a left-click and drag. When selecting an area on high resolution datasets (e.g. BC PRISM), there is the possibility that the user interface will not resolve from screen resolution to dataset resolution. If this is the case, you will be prompted. The can usually be alleviated by zooming the map to a higher zoom level when performing your selection. To adjust the opacity of the climate overlay, there is an opacity slider in the lower left hand corner.
+When in spatial subset selection mode, you can select a rectangular area by doing a left-click and drag. When selecting an area on high resolution datasets (e.g. BC PRISM), there is the possibility that the user interface will not resolve from screen resolution to dataset resolution. If this is the case, you will be prompted. The can usually be alleviated by zooming the map to a higher zoom level when performing your selection.
```
jameshiebert: Excellent. Thanks for making sure that this is documented in the user docs, since the draw icon isn't amazingly obvious.

PR #58, opened 2016-11-08
Feature/download full timeseries

### LINKED ISSUE #28: Add option to web UI to download full time range
As noted in #27, 85% of our raster downloads are for the full timeseries `min_date` to `max_date`. We should add a control to the UI  to make that use case simpler. That could be one of the following:
- add a disable checkbox to the calendar selector to use the full range
- add a hyperlink to the "Download" fieldset that says "Download all" or "Download full time range" or something.

CBMasri:
Fixes #28. The option to download the entire timeseries has been added in the form of a checkbox control to all portals which use a date range. Selecting the checkbox will automatically set the range to the maximum allowed (exact dates vary depending on the dataset).

[Demo](http://docker2.pcic.uvic.ca:8081/downscaled_gcms/map/)

### DIFF
--- a/pdp/static/js/crmp_app.js
+++ b/pdp/static/js/crmp_app.js
@@ -33,18 +33,9 @@ $(document).ready(function () {
     map.composite_filter = '';
     map.getControlsByClass('OpenLayers.Control.DrawFeature')[0].events.register('featureadded', '', pdp.curry(polyChange, map));
 
-    // Specify full timeseries download by setting min/max dates
-    $("#download-full-timeseries").change(
-        function(evt) {
-            if (this.checked) {
-                $("#from-date").datepicker('disable').addClass("disabled").datepicker("setDate", "1870/01/01");
-                $("#to-date").datepicker('disable').addClass("disabled").datepicker("setDate", new Date());
-            } else {
-                $("#from-date").datepicker('enable').removeClass("disabled");
-                $("#to-date").datepicker('enable').removeClass("disabled");
-            }
-        }
-    );
+    // Don't include the full timeseries download option, some of the
+    // stations don't have start/end dates and will get excluded
+    $('#download-all-time').remove()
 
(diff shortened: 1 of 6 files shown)


CBMasri:
I'm wondering if this feature should be removed from the PCDS portal. It looks like some of the stations don't have a min date and get excluded from the results when the checkbox is selected.

pdp/static/js/crmp_app.js
```diff
@@ -33,6 +33,10 @@ $(document).ready(function () {
     map.composite_filter = '';
     map.getControlsByClass('OpenLayers.Control.DrawFeature')[0].events.register('featureadded', '', pdp.curry(polyChange, map));
 
+    // Don't include the full timeseries download option, some of the
+    // stations don't have start/end dates and will get excluded
+    $('#download-all-time').remove()
```
jameshiebert: I find this to be a little non-intuitive. Create a bunch of elements and then manually remove one of them? It's a bit awkward. Maybe a better way would be to include a boolean parameter to `getDateRange()` named `omitFullTimeSeriesCheckbox` (or something less verbose) that's False by default?

pdp/static/js/vic_gen1_app.js
```diff
@@ -72,6 +72,20 @@ $(document).ready(function () {
         }
     );
 
+    // Specify full timeseries download by setting to min/max dates
+    $("#download-full-timeseries").change(
```
jameshiebert: Since this on-change handler is exactly the same as that in canada_ex_app, and (for better or for worse), it doesn't relies on global knowledge (e.g. the element IDs #from-date and #to-date), you could probably just include it once at the end of `getDateRange()`.

pdp/static/js/crmp_app.js
CBMasri: Yes, that makes more sense.

PR #60, opened 2016-11-09
Ensure that the datepicker dates remain consistent with the trange

### LINKED ISSUE #59: Date range displayed doesn't stay consistent with download link on dataset change
There is an inconsistency between the date range displayed on the datepickers and the actual values placed into the download link when switching from one dataset to another.

Steps to reproduce:

1. Navigate to the [downscaled gcm scenarios](http://tools.pacificclimate.org/dataportal/downscaled_gcms/map/) portal.
2. Set the "to" date to the "from" date + 1.
3. Switch to another dataset. "from" date will appear to change to a new `maxDate`.
4. Try to download the new dataset. Date range set in step 2 will still be active.

Possible solutions:

- Refresh the download link to capture the new `maxDate` when changing datasets.
- Prevent the datepicker from updating the values displayed (ie. leave the selected dates as is) on dataset change. This is possibly the better solution, as the introduction of the download-all-timeseries feature will make it easy to reset to the min/max dates.

Affected portals: *hydrologic model output*, *downscaled gcm scenarios*

CBMasri:
Fixes #59. Introduces a simple check to determine if the datepicker start/end dates have been changed from the possible min/max values for the current dataset (default behaviour is to set the displayed dates to the full time range). If either of the dates have been changed, the new value will stay persistent when a new dataset is selected. This ensures that the min/max dates will still be set dynamically according to the ncwms data but the range displayed remains consistent with the `trange` set for the download link.

[Demo](http://docker2.pcic.uvic.ca:8082/hydro_model_out/map/)

### DIFF
--- a/pdp/static/js/pdp_raster_map.js
+++ b/pdp/static/js/pdp_raster_map.js
@@ -135,19 +135,29 @@ function processNcwmsLayerMetadata(ncwms_layer, catalog) {
     });
 }
 
-
 function setTimeAvailable(begin, end) {
     //TODO: only present times available in ncwms capabilities for this layer
     var yearRange = begin.getFullYear().toString(10) + ":" + end.getFullYear().toString(10);
+    var prevMinDate = $(".datepickerstart").datepicker("option", "minDate");
+    var prevMaxDate = $(".datepickerend").datepicker("option", "maxDate");
+    var fromDate = $(".datepickerstart").datepicker("getDate");
+    var toDate = $(".datepickerend").datepicker("getDate");
 
     $.each([".datepickerstart", ".datepickerend"], function (idx, val) {
         $(val).datepicker("option", "minDate", begin);
         $(val).datepicker("option", "maxDate", end);
         $(val).datepicker("option", "yearRange", yearRange);
     });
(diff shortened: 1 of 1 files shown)


jameshiebert:
This approach will actually fail in instances where the available time range is different between the original dataset and the changed-to dataset. For example on the hydro portal:

* Select a GCM (anything from A1B)
* Set the start and end date to be the last day of the timeseries (say 2100-01-01)
* Change the dataset to be something from the historical base
* Notice that the download link will still try to download timestep 54,000 or so, when the available range only goes up to 16k or so.

We might be better off just completely resetting the time range when the user changes datasets

CBMasri:
Hmm yes you are right. It appears that the `begin` and `end` dates being retrieved from the ncwms metadata are not always consistent with the actual values in the dataset. For example, the min/max calendar values for BASE are being set to 1950-2100, even though the available range *should* only be 1950-2006. Unfortunately this means completely resetting the time range will still have the same result, as the max values will still be way out of range. I'm going to dig a little deeper and see if I can get the calendar to set the min/max dates to match those in the dataset.

PR #69, opened 2017-04-26
Improve date handling on raster data portal

### LINKED ISSUE #59: Date range displayed doesn't stay consistent with download link on dataset change

### LINKED ISSUE #64: Datepicker on raster portals ignores calendars
From @ssobie:

> Hi James,
> 
> There appears to be an issue in the BCCAQ data portal when attempting to download a subset of a time series from a downscaled model rather than
> +the full range of data. I've attempted to download 1971-01-01 to 2070-12-30 for HaGEM2-ES for a site in Newfoundland (as per the emails below).
> +The date range I get from the selected file is then: 1971-04-21 to 2072-10-05.
> 
> In the data portal time selection the maximum date available for this model is also 2097-11-05, though if you download the entire series you get
> +the correct range of 1950-01-01 to 2099-12-30.
> 
> 
> This looks like an issue for the other models as well, with the maximum date for 365-day calendar models being 2100-11-25 instead of 2100-12-31. I
> +think the portal must be assuming a gregorian calendar in how it presents the data time.
> 
> Cheers,
> 
> Stephen

corviday:
Added awareness of 360 and 365 day calendars to index-to-date conversion in the CfTime class, to resolve #64. 

Add an event fired to trigger the data download link to refresh its time range when the user switches datasets, to resolve #59. When a user switches to a new dataset, the portal keeps the dates they've previously selected if the newly selected dataset includes them, to make it easier to download matched datasets for comparison.

Test docker [here](http://142.104.230.42:8081/downscaled_gcms/map/)

### DIFF
--- a/pdp/static/js/pdp_raster_map.js
+++ b/pdp/static/js/pdp_raster_map.js
@@ -14,17 +14,7 @@ CfTime.prototype.setMaxTimeByIndex = function (index) {
     this.eDate = this.toDate(index);
     return this.eDate;
 };
-//CfTime.prototype.toDate = function (index) {
-//    if (index === undefined) {
-//        return this.sDate;
-//   }
-//    if (this.units === "days") {
-//        var d = new Date(this.sDate.getTime());
-//        d.setDate(this.sDate.getDate() + index);
-//        d.setDate(d.getDate() + this.getCalendarDayDrift(d));
-//        return d;
-//    }
-//};
+
 CfTime.prototype.toDate = function(index) {
 	if (index === undefined) {
 		return this.sDate;
@@ -97,7 +87,7 @@ function dasToUnitsSince(data) {
         m = reg.exec(s),
         units = m[1],
         dateString = m[3],
-        sDate;
+        sDate;   
     
     var calendar;
     reg = /calendar \"(standard|365_day|360_day)\"/,
@@ -151,7 +141,6 @@ function getNCWMSLayerCapabilities(ncwms_layer) {
 }
 
 function processNcwmsLayerMetadata(ncwms_layer, catalog) {
-
     var layerUrl, maxTimeReq, unitsSinceReq, calendar;
 
     // transform the data_server url into the un-authed catalog based url for metadata
@@ -159,7 +148,7 @@ function processNcwmsLayerMetadata(ncwms_layer, catalog) {
     var reg = /.*\/data\/(.*?)\/.*/g;
     var m = reg.exec(layerUrl);
     layerUrl = layerUrl.replace("data/" + m[1], m[1] + "/catalog")
-
+    
     // Request time variables
     maxTimeReq = $.ajax({
         url: (layerUrl + ".dds?time")
@@ -171,7 +160,6 @@ function processNcwmsLayerMetadata(ncwms_layer, catalog) {
 
     // Process times when both returned
     $.when(maxTimeReq, unitsSinceReq).done(function (maxTime, unitsSince) {
-
         var maxTimeIndex, units, startDate, layerTime;
 
         maxTimeIndex = ddsToTimeIndex(maxTime[0]);
@@ -190,15 +178,42 @@ function processNcwmsLayerMetadata(ncwms_layer, catalog) {
 function setTimeAvailable(begin, end) {
     //TODO: only present times available in ncwms capabilities for this layer
     var yearRange = begin.getFullYear().toString(10) + ":" + end.getFullYear().toString(10);
-
+        
+    //preserve an active range previously set by a user to faciliate downloading matched data.
+    var previousMinimum = $(".datepickerstart").datepicker("option", "minDate");
+    var previousMaximum = $(".datepickerend").datepicker("option", "maxDate");
+    var previousRangeFrom = $(".datepickerstart").datepicker("getDate");
+    var previousRangeTo = $(".datepickerend").datepicker("getDate");
+    
+    //set new maximums and minimums
     $.each([".datepickerstart", ".datepickerend"], function (idx, val) {
         $(val).datepicker("option", "minDate", begin);
         $(val).datepicker("option", "maxDate", end);
         $(val).datepicker("option", "yearRange", yearRange);
     });
-    $(".datepicker").datepicker("setDate", begin);
-    $(".datepickerstart").datepicker("setDate", begin);
-    $(".datepickerend").datepicker("setDate", end);
+
+    //try to keep the active range, if it was specified and is possible.
+    //fall back to the beginning and end of the new dataset.
+    if(previousMinimum 
+    		&& (previousMinimum.getTime() != previousRangeFrom.getTime()) 
+    		&& (previousRangeFrom.getTime() >= begin.getTime() )) {
+    	$(".datepickerstart").datepicker("setDate", previousRangeFrom);
+    	$(".datepicker").datepicker("setDate", previousRangeFrom);     	
+    } else {
+    	$(".datepickerstart").datepicker("setDate", begin);
+        $(".datepicker").datepicker("setDate", begin);
+    }
+    
+    if(previousMaximum 
+    		&& (previousMaximum.getTime() != previousRangeTo.getTime()) 
+    		&& (previousRangeTo.getTime() <= end.getTime() )) {
+    	$(".datepickerend").datepicker("setDate", previousRangeTo);
+    } else {
+    	$(".datepickerend").datepicker("setDate", end);
+    }
+    
+    //fire a change event to trigger the download link to update
+   $("[class^='datepicker']").trigger("change");
 }
 
 function intersection(b1, b2) {
--- a/pdp/static/js/pdp_raster_map.js
+++ b/pdp/static/js/pdp_raster_map.js
@@ -100,7 +100,7 @@ function dasToUnitsSince(data) {
     var calendar;
     reg = /calendar \"(standard|gregorian|365_day|noleap|360_day)\"/,
     m = reg.exec(s),
-    calendar = m[1];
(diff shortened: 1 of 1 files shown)


pdp/static/js/pdp_raster_map.js
```diff
@@ -53,25 +87,31 @@ function dasToUnitsSince(data) {
         m = reg.exec(s),
         units = m[1],
         dateString = m[3],
-        sDate;
-
+        sDate;   
+    
+    var calendar;
+    reg = /calendar \"(standard|365_day|360_day)\"/,
```
jameshiebert: Ensure that your code also supports the other valid aliases for standard and 365_day that are described [in section 4.4.1 of this document](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch04s04.html). In particular, make sure that you support `noleap` and `gregorian`. I don't think that we need to support `all_leap`, `proleptic_gregorian`, or `julian` at this time.

jameshiebert: Also, if calendar comes in empty (it *shouldn't*, but just in case), you should make sure that you default to 'standard'.

pdp/static/js/pdp_raster_map.js
```diff
@@ -101,15 +141,14 @@ function getNCWMSLayerCapabilities(ncwms_layer) {
...
-    var layerUrl, maxTimeReq, unitsSinceReq;
+    var layerUrl, maxTimeReq, unitsSinceReq, calendar;
 
     // transform the data_server url into the un-authed catalog based url for metadata
     layerUrl = catalog[getNcwmsLayerId(ncwms_layer)];
     var reg = /.*\/data\/(.*?)\/.*/g;
     var m = reg.exec(layerUrl);
     layerUrl = layerUrl.replace("data/" + m[1], m[1] + "/catalog")
-
+
```
jameshiebert: Drop this added whitespace.

pdp/static/js/pdp_raster_map.js
```diff
@@ -53,25 +87,31 @@ function dasToUnitsSince(data) {
...
+    m = reg.exec(s),
+    calendar = m[1];
+        
     reg = /(\d{4})-(\d{1,2})-(\d{1,2})( |T)(\d{1,2}):(\d{1,2}):(\d{1,2})/g;
     m = reg.exec(dateString);
     if (m) {
         sDate = new Date(m[1], parseInt(m[2], 10) - 1, // Months in das result are 1-12, js needs 0-11
                          m[3], m[5], m[6], m[7], 0);
-        return [units, sDate];
+        return [units, sDate, calendar];
```
jameshiebert: Yeah, this works to add another item to the list. However, if we end up hacking in this code *much* more than this, we'll probably want to convert the return value to be a properly attributed key/value mapping (rather than having a list of items that have a particular meaning without any context).

pdp/static/js/pdp_raster_map.js
```diff
@@ -4,38 +4,72 @@
...
-        var msPerDay = 1000 * 60 * 60 * 24,
-            msDiff = d.getTime() - this.sDate.getTime(),
-            days = msDiff / msPerDay;
-        return Math.floor(days);
+    var days;
+    var msPerDay = 1000 * 60 * 60 * 24;
+    if(this.units == "days") {
+    	if(this.calendar != "standard") {
+    		var daysPerYear = this.calendar=="365_day" ? 365 : 360;
+    		days = (d.getFullYear() - this.sDate.getFullYear()) * daysPerYear;
```
jameshiebert: The two conditionals above will need a rewrite.

pdp/static/js/pdp_raster_map.js
```diff
@@ -4,38 +4,72 @@
...
+    		var steppedDate = new Date(d);
```
jameshiebert: What does "stepped"Date mean?

pdp/static/js/pdp_raster_map.js
```diff
@@ -4,38 +4,72 @@
...
-
-    if (this.units === "days") {
-        var msPerDay = 1000 * 60 * 60 * 24,
-            msDiff = d.getTime() - this.sDate.getTime(),
-            days = msDiff / msPerDay;
-        return Math.floor(days);
+    var days;
+    var msPerDay = 1000 * 60 * 60 * 24;
+    if(this.units == "days") {
+    	if(this.calendar != "standard") {
```
jameshiebert: Doesn't necessarily matter, but I'd probably flip the logic here. If it *is* a standard calendar, do the simplest and normal thing first. And if it's not, then go into the weird exception cases. It's functionally the same, but I think it reads a little better.

pdp/static/js/pdp_raster_map.js
```diff
@@ -139,15 +178,42 @@ function processNcwmsLayerMetadata(ncwms_layer, catalog) {
...
         $(val).datepicker("option", "maxDate", end);
         $(val).datepicker("option", "yearRange", yearRange);
     });
-    $(".datepicker").datepicker("setDate", begin);
-    $(".datepickerstart").datepicker("setDate", begin);
-    $(".datepickerend").datepicker("setDate", end);
+
+    //try to keep the active range, if it was specified and is possible.
+    //fall back to the beginning and end of the new dataset.
+    if(previousMinimum
```
jameshiebert: Question: under what circumstances would `previousMinimum` be falsey? Would it be NULL or undefined and is it worth making that comparison explicit?

pdp/static/js/pdp_raster_map.js
```diff
@@ -139,15 +178,42 @@ function processNcwmsLayerMetadata(ncwms_layer, catalog) {
...
+    if(previousMaximum 
+    		&& (previousMaximum.getTime() != previousRangeTo.getTime()) 
+    		&& (previousRangeTo.getTime() <= end.getTime() )) {
+    	$(".datepickerend").datepicker("setDate", previousRangeTo);
+    } else {
+    	$(".datepickerend").datepicker("setDate", end);
+    }
+    
+    //fire a change event to trigger the download link to update
+   $("[class^='datepicker']").trigger("change");
```
jameshiebert: Good, that makes sense!

pdp/static/js/pdp_raster_map.js
corviday: `previousMinimum` is `null` during the initial page load, the first time through this function.

pdp/static/js/pdp_raster_map.js
corviday: It's the same date as the date-of-interest (the date we're looking for the index of) but "stepped backwards" a particular number of whole years. That is, if we're looking for the index of April 1 1970, and our start date is January 1 1950, the stepped-back date is April 1 1950. It's used to calculate the "remainder" days in the index that arise if the start date and the date-of-interest are different days of the year.

Since it's used to calculate ` remainderDays`, I renamed it `remainderDate`, hopefully that will be a little clearer.

pdp/static/js/pdp_raster_map.js
```diff
@@ -55,18 +97,23 @@ function dasToUnitsSince(data) {
         dateString = m[3],
         sDate;
 
+    var calendar;
+    reg = /calendar \"(standard|gregorian|365_day|noleap|360_day)\"/,
+    m = reg.exec(s),
+    calendar = m[1];
```
jameshiebert: Last change request, I swear! Line 103 will fail (and stop the script) in the event that the DAS responds with a calendar that is not one of those listed. E.g.:

```javascript
> var s = "calendar \"foo\"";
undefined
> reg = /calendar \"(standard|gregorian|365_day|noleap|360_day)\"/,
... m = reg.exec(s),
... calendar = m[1];
TypeError: Cannot read property '1' of null
    at repl:3:13
    at realRunInThisContextScript (vm.js:22:35)
    at sigintHandlersWrap (vm.js:98:12)
    at ContextifyScript.Script.runInThisContext (vm.js:24:12)
    at REPLServer.defaultEval (repl.js:346:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:545:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:188:7)
```
Please ensure that a default value (standard) is assumed here as well and be robust against possible unknown input.

corviday: I considered adding a similar default value for "units" in the same function, but wasn't entirely sure whether "days" was a good default.

pdp/static/js/pdp_raster_map.js
```diff
@@ -100,7 +100,7 @@ function dasToUnitsSince(data) {
     var calendar;
     reg = /calendar \"(standard|gregorian|365_day|noleap|360_day)\"/,
     m = reg.exec(s),
-    calendar = m[1];
+    calendar = m ? m[1] : "standard";
```
jameshiebert: Cool, that should work.

PR #70, opened 2017-05-04
Add a test and documentation for Zip64 encoding.

corviday:


jameshiebert:
I've pushed a new version (0.7) of pydap.responses.aaigrid [to our PyPI](https://pypi.pacificclimate.org/simple/pydap-responses-aaigrid/). So how about you go ahead and bump the requirements and remove the xfail mark?

jameshiebert:
Looks good!

Issue #37, opened 2016-01-18
Raster portals don't support multiple grid resolutions on the same page


mschnorb:
When attempting to download precipitation data from Gridded Hydrologic Model Output data portal, the resultant netCDF file only serves data over a truncated spatial domain of 47.5N to 57.7N by -139W to -125.5W (see image below), when data for all of BC are expected.

![precip_hydro-portal](https://cloud.githubusercontent.com/assets/9890345/12403401/72d638e6-bde8-11e5-8d8b-6bb6c4513dde.png)

Steps taken:
1. Select HadCM A1B
2. Select Precipitation variable
3. Set date range from 1950/01/01 to 1979/12/31
4. Output format NetCDF (default)
5. Click download

Expected precipitation data for the selected date range covering all of BC (as shown in the portal map preview). However, the resultant netCDF file only provides precipitation over on a truncated spatial array (lat=164, lon=216) as opposed to all of BC.Note that the same problem was encountered when attempting to download precipitation for the GFDL A1B scenario.

Time: 18/01/2016 1:32 pm (PST)
IP address: 142.104.194.119

jameshiebert:
Thanks for the detailed bug report. This *is* actually a known issue, but apparently no one has written it up on the issue tracker, so thanks for doing so.

The root of the problem is that we have several variables on this page with different spatial domains / grid properties. And when we designed the page, we unfortunately assumed that all of the listed datasets on a given page would have the same spatial properties. We didn't check this assumption when adding the forcings data to this page, so essentially, you will see this issue with any of the forcings data sets (precip, tasmin, tasmax).

If you really just want to download the whole domain, there's an easy temporary work around (until we fix the problem).

1. Select everything that you want (dataset, time range, output format) *except* for the spatial domain
1. Copy/paste the download link (e.g. http://tools.pacificclimate.org/dataportal/data/hydro_model_out/pr+tasmin+tasmax+wind_day_CSIRO35_A2_run1_19500101-21001231.nc.nc?pr[0:54787][0:163][0:215]&)
1. Omit the grid parameters at the end of the link. E.g. replace ``[0:163][0:215]`` with ``[:][:]``
1. Use the edited link to download the full spatial domain

mschnorb:
Thanks James. ~MS

PR #72, opened 2017-12-13
Add Gridded Observations portal

corviday:
Add a new portal for gridded observational / historical data. ~~This portal requires the user to designate an area for download using the map, which is an end-run around issue #37 . To get the entire spatial domain, you'd draw a rectangle around the whole thing.~~

[Docker for the new portal](http://docker-dev01.pcic.uvic.ca:30032/gridded_observations/map/)

### DIFF
--- a/doc/source/pdp.rst
+++ b/doc/source/pdp.rst
@@ -31,3 +31,6 @@ The pdp module
 
 .. automodule:: pdp.portals.vic_gen1
    :members:
+
+.. automodule:: pdp.portals.gridded_observations
+   :members:
--- a/pdp/portals/gridded_observations.py
+++ b/pdp/portals/gridded_observations.py
@@ -1,0 +1,60 @@
+'''The pdp.portals.gridded_observations module configures a raster portal
+ which serves gridded climate data used by the VIC model.'''
+
+from pdp import wrap_auth
+from pdp.dispatch import PathDispatcher
+from pdp_util.map import MapApp
+from pdp_util.raster import RasterServer, RasterCatalog, RasterMetadata
+from pdp_util.ensemble_members import EnsembleMemberLister
+
+from pdp.minify import wrap_mini
+from pdp.portals import updateConfig, raster_conf
+
+ensemble_name = 'gridded-obs-met-data'
+url_base = 'gridded_observations'
+
+class GriddedObservationsEnsembleLister(EnsembleMemberLister):
+
+    def list_stuff(self, ensemble):
+        dataset_names = {
+            "ANUSPLIN_CDA_v2012.1": "ANUSPLIN",
+            "SYMAP_BC_v1": "VIC FORCINGS",
+            "TPS_NWNA_v1": "TPS"}
+        for dfv in ensemble.data_file_variables:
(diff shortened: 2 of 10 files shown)


doc/source/pdp.rst
```diff
@@ -31,3 +31,6 @@ The pdp module
 
 .. automodule:: pdp.portals.vic_gen1
    :members:
+
```
jameshiebert: Way to keep the docs in sync!

pdp/static/js/pdp_controls.js
```diff
@@ -308,6 +310,11 @@ RasterDownloadLink.prototype = {
 
     // Take the URL template and substitute each of the desired state variables
     getUrl: function () {
+        //if this portal requies the user to select an area, don't generate a URL
+        //unless they've made a selection.
+        if(this.requireArea && !this.areaSpecified) {
```
jameshiebert: I'm not a big fan of this solution. I'd rather that we omit the i/j parameters by default (which is a valid OPeNDAP request) until the user selects a rectangle.

This map require some changes to the `RasterDownloadLink` in `pdp_controls.js` or simply initializing the `RasterDownloadLink` with empty coordinates.

pdp/portals/gridded_observations.py
```diff
@@ -0,0 +1,61 @@
...
+from pdp.portals import updateConfig, raster_conf
+
+ensemble_name = 'gridded-obs-met-data'
+url_base = 'gridded_observations'
+
+class GriddedObservationsEnsembleLister(EnsembleMemberLister):
+
+    def list_stuff(self, ensemble):
+        dataset_names = {
+            "ANUSPLIN_CDA_v2012.1": "ANUSPLIN",
```
jameshiebert: We'll have to wait to hear back from Hydro, but I'd like to omit ANUSPLIN and SYMAP until #37 is fixed.

pdp/static/js/gridded_observations_app.js
```diff
@@ -0,0 +1,85 @@
...
+    catalogUrl = "../catalog/catalog.json";
+    catalog_request = $.ajax(catalogUrl, {dataType: "json"});
+
+    capabilities_request = getNCWMSLayerCapabilities(ncwmsLayer);
+
+    document.getElementById("pdp-controls").appendChild(getObsControls(pdp.ensemble_name));
+    document.getElementById("pdp-controls").appendChild(getRasterDownloadOptions(true));
+
+    // Data Download Link
+    dlLink = new RasterDownloadLink($('#download-timeseries'), ncwmsLayer, undefined, 'nc', 'sm', '0:54787', '0:163', '0:215', true);
```
jameshiebert: Here, you may want to change '0:163' to '' and '0:215' to ''.

corviday:
Here is what I hope is the last test docker on this project. It is built on [this branch of pydap-pdp](https://github.com/pacificclimate/pydap-pdp/pull/2), and initializes spatial indices on the Gridded Observations and Vic Gen1 portals to `[][]`:

http://docker-dev01.pcic.uvic.ca:30032/gridded_observations/map/

It all seems to be working, as far as I can tell! 

It's not a perfect systems / integration test: it's using an ncWMS2 instance running on docker-dev01 (that only contains the Gridded Observation files) instead of the ncWMS-PCIC instance on pizza.

PR #80, opened 2018-05-23
Updates the display titles of gridded observation portal and its datasets

corviday:
I've requested proofreading and will merge after that's done.


jameshiebert:
Looks good to me!

PR #82, opened 2018-05-23
Extend map layer to include entire PNWNAmet dataset

corviday:
largerarea
Extends the map and layer bounds to longitude -170 to display the PNWNAmet dataset, which goes from -168.96875 to -101.03125.


jameshiebert:
Perfect! Glad that it didn't go past -180 or we would have been hosed.

PR #91, opened 2018-08-23
Resolve #90; add defaults for undefined indices

### LINKED ISSUE #90: Undefined end for time slice
From the logs:
npkfndadjgbfcbkb

Several requests for download data with time ranges of the form [0:undefined], which unsurprisingly errored out. The invalid time range is presumably generated by the front end's [RasterDownloadLink](https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/pdp_controls.js#L392) class.

RasterDownloadLinks are [initially created](https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/gridded_observations_app.js#L26) with the constant valid time range `[]`, which is unlikely to be the source of the problem, so the issue is probably happening with `RasterDownloadLink.onTimeChange()`. This function gets dates from the datepicker and converts them into indices.

I've fooled around with the datepicker and link generation functions, and I haven't been able to recreate this bug. However, adding a quick check to make sure that `start` and `end` aren't `undefined` and a falling back to the beginning or end of the time range if they are should result in no change under normal circumstances and a more graceful failure (providing more information than the user wanted, instead of a 500 error) if this bug recurs.

corviday:
Addresses #90, where the front end at least once has generated download links with time slices of the form `[0:undefined]`. I wasn't able to reproduce the error, but I added a check for it: if the starting or ending value of a time slice is undefined, a default value (0 for start, Null for end) will be substituted.

Should the error recur, instead of generating an invalid link that returns no data, the front end should now generate a link that, at worst, returns more data than the user wanted, since it defaults to the entire time range.

### DIFF
--- a/pdp/static/js/pdp_controls.js
+++ b/pdp/static/js/pdp_controls.js
@@ -397,6 +397,10 @@ RasterDownloadLink.prototype = {
         end = $(".datepickerend").datepicker("getDate");
         end = this.layer.times.toIndex(end);
 
+        //if either start or end is undefined, fall back to the full time range
+        start = start === undefined ? 0 : start;
+        end = end === undefined ? "" : end;
+
         this.trange = start + ':' + end;
         this.trigger();
     }


jameshiebert:
Looks good to me! Feel free to merge this once the push check finishes.

corviday:
Odd. When I go to the [travis check's page](https://travis-ci.org/pacificclimate/pdp/builds/419814669?utm_source=github_status&utm_medium=notification) to see what's holding it up, it appears to be done, but it's showing as still in process on this PR.

jameshiebert:
Well, merge away then :)

PR #99, opened 2019-01-25
Upgrade requirements

jameshiebert:
I did a bit of work on upgrading all of the available requirements for `pdp`. *Most* of the work here, is actually hidden in the [new version of `pdp_util`](https://github.com/pacificclimate/pdp_util/compare/0.2.8...master). To upgrade `PyCDS` I had to rework a bunch of the session management testing code (in an unfortunate way, but that's a different story).

As a bonus, I found a couple of places in `pdp_util` where the database sessions were being used outside of a context manager resulting in zombie database connections in transaction. Those are cleaned up now, so we should stop leaking connections.

In the first draft of this PR, there are an insane number of warnings generated by running the test suite (lots of sketchy `numpy` and `h5py` issues that could affect the numeric results). I've removed that patch for now so that we can focus on the easy (and value adding) things, and work on the numeric upgrades at a later date.

### DIFF
--- a/pdp/portals/pcds.py
+++ b/pdp/portals/pcds.py
@@ -10,6 +10,7 @@ from pdp_util.map import MapApp
 from pdp.minify import wrap_mini
 from pdp.portals import updateConfig
 
+from pdp_util import session_scope
 from pdp_util.counts import CountStationsApp, CountRecordLengthApp
 from pdp_util.legend import LegendApp
 from pdp_util.agg import PcdsZipApp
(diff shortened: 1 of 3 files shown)


jameshiebert:
I think that I'm going to remove all of the non-essential package upgrades (especially numpy and h5py) in favor of doing the `pdp_util` upgrade and getting the leaking database sessions easily merged. Stand by...

jameshiebert:
@corviday I think this is good to go if you want to take a look at it.

PR #114, opened 2019-04-05
Adds a FAQ entry to the docs about slow timeseries

jameshiebert:
Adds a FAQ entry that discusses the issue with time vs. space major
array ordering, data fragmentation, what type of performance users can
expect, and what to do about it.


jameshiebert:
Cool, I'll merge it then.

PR #115, opened 2019-05-04
Fix front-end time range selection

### LINKED ISSUE #110: Off-by-one error in front-end time range code
The front-end code uses OPeNDAPs DDS response to determine the length of the time dimension for each dataset.

Unfortunately, there appears to be an off-by-one error. The DDS response returns the *length* of the time dimension, and then the front-end code [naively uses](https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/pdp_raster_map.js#L88) the length as the final raster index.

However, in the OPeNDAP spec [array indices are inclusive](https://docs.opendap.org/index.php/DAP4:_Specification_Volume_1#Array_Subsetting_in_Index_Space) at both the start and stop of the range. So the ending index should be `length - 1`, I believe.

Not sure how we didn't run into this before... maybe the client would just close the connection when it received all of the data?

rod-glover:
Resolves #110 

This PR is much too big. Fortunately it can, and will, be broken into several neat and largely independent pieces, as follows:

1. Nodejs and test infrastructure.
1. ES5 support for maintainable classes.
1. New calendar classes.
1. Modifications to the PDP code.
1. Probably something else I am forgetting right now.

Also reassuring to the possibly appalled prospective reviewer is the fact that most of the 13,000 lines of new/modified code is consumed by a `package-lock.json` and by a `lodash` core (the latter will be included in the new PDP app code). The actual amount of reviewable code is much much smaller than this.


pdp/static/js/pdp_raster_map.js
```diff
@@ -1,223 +1,157 @@
...
+        return new calendars.CfDatetime(newSystem, fallbackDate);
     }
-
-    if(previousMaximum
-            && (previousMaximum.getTime() != previousRangeTo.getTime())
-            && (previousRangeTo.getTime() <= end.getTime() )) {
-        $(".datepickerend").datepicker("setDate", previousRangeTo);
-    } else {
-        $(".datepickerend").datepicker("setDate", end);
+    if (_.isEqual(date.cfTimeSystem, newSystem)) {
```
rod-glover: date.system?

pdp/static/js/pdp_raster_map.js
```diff
@@ -252,12 +186,28 @@ function getRasterBbox(capabilities, layer_name) {
...
-    //  units = ncwms_layer.times.units,
-        t0i = ncwms_layer.times.toIndex(t0),
-        tni = ncwms_layer.times.toIndex(tn);
-    return [t0i, tni];
+    // var t0 = $(".datepickerstart").datepicker("getDate"),
+    //     tn = $(".datepickerend").datepicker("getDate"),
+    // //  units = ncwms_layer.times.units,
+    //     t0i = ncwms_layer.times.toIndex(t0),
+    //     tni = ncwms_layer.times.toIndex(tn);
+    // return [t0i, tni];
```
rod-glover: Remove

pdp/static/js/pdp_raster_map.js
```diff
@@ -252,12 +186,28 @@ function getRasterBbox(capabilities, layer_name) {
...
+
+    function toCfTime(string) {
+        var datetime = calendars.SimpleDatetime.fromLooseFormat(string);
+        var t0 = calendars.CfTime.fromDatetime(
```
rod-glover: return

How does this even work??? Add a test that would reveal this error.

rod-glover:
At this point, I'm going to call this good. We have working apps, working tests, and CI.

Now to break it up into some sane PRs.

rod-glover:
Superseded by #116

PR #116, opened 2019-05-13
Fix front-end time range selection

### LINKED ISSUE #110: Off-by-one error in front-end time range code

### LINKED ISSUE #127: Disable download buttons when date range is invalid
Currently, user can enter an invalid date, then trigger a download, and that results in a 500 error, and total loss of context. Nasty.

rod-glover:
Resolves #110 
Resolves #127 

This branch/PR is the "master" PR for a series of smaller PRs that break up the mondo PR #115.

Once all sub-PRs are merged into this PR, it can be finally vetted and merged into master. This is safer and easier (because effectively a sandbox) than merging the component PRs piecemeal directly into master.

Plan for component PRs:

1. Node.js, test infrastructure (Jest), and ES5 support for maintainable classes.
1. New calendar classes.
1. Modifications to the PDP code.


README.md
```diff
@@ -127,10 +127,23 @@ pyenv/bin/py.test -vv --tb=short tests
...
+First set up the environment variables that do not default to usable values. 
+Obtain the user ID's and passwords necessary for the two databases from PCIC IT.
+We typically use port 8000 but any port will do.
+
+```bash
+export DSN=postgresql://<USER>:<PASSWORD>@db3.pcic.uvic.ca/pcic_meta
+export DATA_ROOT=http://127.0.0.1:<PORT>/data
+export PCDS_DSN=postgresql://<USER>:<PASSWORD>@db3.pcic.uvic.ca/crmp
+export APP_ROOT=http://127.0.0.1:<PORT>
+```
```
corviday: This is probably outside the scope of this PR, but I feel there should be a better way to organize the README. There's a lot of information in it, and it's not intuitively organized. There's a section on environmental variables that describes them in general terms, and then down here there's specific options for them. 

I remember starting with this project and being completely unsure which of the many instructions in various sections applied to what I was trying to do.

jameshiebert: https://github.com/pacificclimate/pdp/issues/118

rod-glover:
@jameshiebert , @corviday,  this PR is finally ready to merge. All the tributary PRs have passed review and testing.

For your entertainment, I've put up a [demo](http://docker-dev01.pcic.uvic.ca:30509).

corviday:
Likely an issue for another PR: if you select some stations on the PCDS portal by drawing a polygon, it would be nice to have the start and end dates adjust to reflect data actually present with those stations. The previous version of the portal doesn't do this either, though, and if the user selects stations and then selects dates not present on those stations, the stations disappear from the map, which is a solid UI indication of what's going on, so I don't see any urgency to this.

corviday:
Also a seperate issue, not introduce by this PR nor very urgent: nothing requires your end date to be >= than your start date or displays an error message if you select one that's earlier.

rod-glover:
See #136

rod-glover:
Are we approaching LGTM time yet?

corviday:
LGTM.

jameshiebert:
I'm happy w/ it. Let's go for it!

PR #119, opened 2019-05-17
Add calendar classes

rod-glover:
This PR adds calendar classes. They are divided into the following groups, with subclassing shown by indentation:

* `SimpleDatetime`
* `Calendar`
   * `GregorianCalendar`
   * `Fixed365DayCalendar`
   * `Fixed360DayCalendar`
* `CalendarFactory`
* `CalendarDatetime`
* `CfTimeSystem`
* `CfDatetime`

The signatures of many of the methods, including constructors, that take datetime inputs are not optimal. Some take `SimpleDatetime`, (most) others take the components `year`, `month`, etc. There are also several other places where signatures could be improved, particularly in constructors. But this is workable, if a little rough on the programmer who has constantly to refer to the documentation to remember what the hell arguments to provide to a given method.

I invite feedback on all of this.

Note: The base branch for this PR is i110-main; see #116.

### DIFF
--- a/pdp/static/js/__test__/calendars.test.js
+++ b/pdp/static/js/__test__/calendars.test.js
@@ -1,0 +1,877 @@
+var each = require('jest-each').default;
+
+require('./globals-helpers').importGlobals([
+    { module: 'js/condExport', name: 'condExport' },
+    { module: 'js/classes.js', name: 'classes' },
+    { module: 'js/calendars.js', name: 'calendars' },
+], '../..');
+
+var formatDatetimeRaw = calendars.formatDatetimeRaw;
+var formatDatetimeISO8601 = calendars.formatDatetimeISO8601;
+var SimpleDatetime = calendars.SimpleDatetime;
+var Calendar = calendars.Calendar;
+var GregorianCalendar = calendars.GregorianCalendar;
+var Fixed365DayCalendar = calendars.Fixed365DayCalendar;
+var Fixed360DayCalendar = calendars.Fixed360DayCalendar;
+var CalendarFactory = calendars.CalendarFactory;
+var CalendarDatetime = calendars.CalendarDatetime;
+var CfTimeSystem = calendars.CfTimeSystem;
+var CfDatetime = calendars.CfDatetime;
+
+var epochYear = 1900;
+var gregorianCalendar = new GregorianCalendar(epochYear);
+var fixed365DayCalendar = new Fixed365DayCalendar(epochYear);
+var fixed360DayCalendar = new Fixed360DayCalendar(epochYear);
+
+// TODO: Add tests for untested methods. These are all mostly shortcuts and
+// convenience methods.
+
+
+// A true universal, upon which we all can agree.
+var msPerDay = 1000 * 60 * 60 * 24;
+
+function leapDaysSinceEpoch(calendar, year) {
+    // Up and including `year`.
+    // Valid only up to 2099!!!
+    return Math.floor((year - calendar.epochYear) / 4);
+}
+
+
+describe('formatDatetimeRaw', function () {
+    each([
+        [1900, 1, 2, 3, 4, 5, '1900-1-2T3-4-5'],
+        [undefined, 1, 2, 3, 4, 5, 'undefined-1-2T3-4-5'],
+    ]).it('%#',
+        function (year, month, day, hour, minute, second, expected) {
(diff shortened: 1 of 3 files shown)


pdp/static/js/__test__/calendars.test.js
```diff
@@ -0,0 +1,896 @@
...
+var Calendar = calendars.Calendar;
+var GregorianCalendar = calendars.GregorianCalendar;
+var Fixed365DayCalendar = calendars.Fixed365DayCalendar;
+var Fixed360DayCalendar = calendars.Fixed360DayCalendar;
+var CalendarFactory = calendars.CalendarFactory;
+var CalendarDatetime = calendars.CalendarDatetime;
+var CfTimeSystem = calendars.CfTimeSystem;
+var CfDatetime = calendars.CfDatetime;
+
+var epochYear = 1900;
```
corviday: The PDP includes a few datasets that start at 1850. Will this be an issue?

corviday: Actually, upon further investigation, the PDP only includes ONE timeset that starts then, which is probably a mistake, and I should investigate it. But, hypothetically, would we only be able to include post-1900 timesets in the future?

pdp/static/js/calendars.js
```diff
@@ -0,0 +1,807 @@
...
+        createCalendar: function(type) {
+            // Note we do not presently make any distinction between standard,
+            // Gregorian, and proleptic Gregorian calendars. This is actually wrong,
+            // but it will do.
+            switch (type) {
+                case 'standard':
+                case 'gregorian':
+                case 'proleptic_gregorian':
+                    return new GregorianCalendar(this.epochYear);
+                case '365_day':
```
corviday: May also need to support the `noleap` synonym for`'365_day`

corviday:
I'm having a hard time understanding this PR. It might just be due to not seeing how it fits together with other components. Could you perhaps talk a bit about how it is intended to be used, or post a code snippet?

pdp/static/js/__test__/calendars.test.js
rod-glover: This is a test. The default epoch year is 1800, for that very reason. But it is good to test with some other year.

rod-glover:
@corviday See the new section in the README for explanations and code snippets.

rod-glover:
It might make sense to break the calendars module up into two separate modules, as follows:

- `Calendar` classes, `CalendarDatetime`
- `CfTimeSystem` and `CfDatetime`

But in this project at least, they are all used together.

pdp/static/js/calendars.js
```diff
@@ -0,0 +1,818 @@
...
+
+    var isInteger = Number.isInteger;
+
+    var isArray = Array.isArray;
+
+    function isInRange(v, min, max) {
+        return min <= v && v < max;
+    }
+
+    function format(places, num) {
```
corviday: Seems like you could have some weird results with decimals.
```javascript
> format(3, .967)
< "967"
```

Will this ever be used to format values from the time variable of a netCDF file? They are not always whole numbers.

```
netcdf pr_day_BCCAQv2+ANUSPLIN300_bcc-csm1-1-m_historical+rcp85_r1i1p1_19500101-21001231 {
dimensions:
    lon = 1068 ;
    lat = 510 ;
    time = 55115 ;
variables:
    double lon(lon) ;
    double lat(lat) ;
    double time(time) ;
        time:units = "days since 1850-01-01 00:00:00" ;
        time:calendar = "365_day" ;
        time:long_name = "Time" ;
        time:standard_name = "Time" ;
    short pr(time, lat, lon) ;
data:
 
 time = 36500.5, 36501.5, 36502.5, 36503.5, 36504.5, 36505.5, 36506.5,
...
}
```

pdp/static/js/calendars.js
```diff
@@ -0,0 +1,818 @@
...
+        }
+    }
+
+
+    function formatDatetimeRaw(year, month, day, hour, minute, second) {
+        // Return a string representing the argument values in an
+        // ISO 8601-like format, but without any checking or fancy formatting.
+        // Useful for error messages.
+        return '' + year + '-' + month + '-' + day +
+            'T' + hour + '-' + minute + '-' + second;
```
corviday: Should it be colons between hours, minutes, and seconds?

PR #121, opened 2019-05-24
Update canada_ex_app (Downscaled GCM Scenarios)

rod-glover:
This PR updates `canada_ex_app`, which provides the Downscaled GCM Scenarios portals. 

This is the main and most complicated app where changes involving date-selection must be done. (Others, simpler, to come in a later PR.)

Overview

- The philosophy of this update is to disturb the existing code as little as possible. Dates chosen with the date-pickers were and continue to be stored in the DOM, rather than in application-state variables as most modern approaches do.

- To verify that the intended changes have been correctly introduced, the app has been packaged so that it can be tested. The tests run the app in a simulated browser (`jsdom`). Tests use jQuery to examine the the DOM to make sure that it contains what it should. This can include dynamic behaviour, e.g, before and after user interaction with the UI (simulated in test code by diddling the DOM with jQuery).

- To support testing, all files now export their contents using `condExport`. This is a no-op as far as functionality is concerned.

- Never fear: 90,000 of the lines added are library modules, not for review. That still leaves about 1500 LOC for review, but most of that is test code.

Changes to the app

- Main changes are in the following files:
  - `pdp/static/js/data-services.js`: Module extracting calls to data services (backend). This reduces redundancy in the code (of which there was a fair bit) and makes it much simpler to mock these services for unit testing.
  - `pdp/static/js/pdp_controls.js`: Modifies date-pickers to use new calendar code, and (unfortunately) not to use a cool date-picker UI element (those only work with native JS Date, which is only Gregorian calendar). 
  - `pdp/static/js/pdp_raster_map.js`: Modifies layer code to use new calendar code. This is where metadata is transformed into `CfTimeSystem` and related computations of `CfDatetime` values are done.

- Less substantial changes to the app:
  - `pdp/static/js/canada_ex_app.js`: Wrap the app as a function so that it can be tested.
  - `pdp/static/js/canada_ex_map.js`: Use data-services module.
  - `pdp/static/css/controls.css`: Layout and formatting for new date-picker controls.
  - `pdp/__init__.py`: Serve the new static files.

- `condExport` only (repackages the contents so they are usable in tests):
  - `pdp/static/js/pdp_download.js`
  - `pdp/static/js/pdp_filters.js`
  - `pdp/static/js/pdp_map.js`
  - `pdp/static/js/pdp_vector_map.js`

Tests and test support

- Tests: Tests of updated app code:
  - `pdp/static/js/__test__/canada_ex_app.test.js`: Applies `date-filter-tests` to `canada_ex_app`.
  - `pdp/static/js/__test__/date-filter-tests.js`: Main body of tests for apps. Tests date-filter UI. Abstracted out so it can be applied to other apps (to come in future PR(s)).
  - `pdp/static/js/__test__/pdp_controls.test.js`: Tests PDP controls, specifically download link content and behaviour.
  - `pdp/static/js/__test__/pdp_raster_map.test.js`: Tests map raster layer code as it relates to date picking.

- Test support:
  - `pdp/static/js/OL/OpenLayers-2.13.1.debug.js`: Used for tests to enable debugging of OL-related code. Not used in production app.
  - `pdp/static/js/__mocks__/data-services.js`: Mocks for data services.
  - `pdp/static/js/__test__/mock-helpers.js`: Helpers for building mocks.
  - `pdp/static/js/__test__/app-test-helpers.js`: You guessed it. Small and simple.


rod-glover:
Suggestion for reviewers: In GitHub PR, on Files changed tab, select the gear icon, then select "Hide whitespace changes". It reduces the noise in the diffs.

rod-glover:
I will shortly put up a Docker instance running the modified code.

rod-glover:
[Demo version](http://docker-dev02.pcic.uvic.ca:30509/downscaled_gcms/map/) now running.

Only the Downscaled GCMs app has been updated. Other apps are probably erroneous now.

corviday:
I'm looking at the demo. Is this update solving issue #110 in the downscaled gcms scenarios portal, or is it a step toward that end?

The URL I see when I hover over the download link the portal upon first opening it is https://services.pacificclimate.org/dataportal/data/downscaled_gcms/tasmax_day_BCCAQv2+ANUSPLIN300_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc.nc?tasmax[0:55152][0:510][0:1068]&  but this file has 55115 timesteps:

```bash
ncdump -h tasmax_day_BCCAQv2+ANUSPLIN300_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc 
netcdf tasmax_day_BCCAQv2+ANUSPLIN300_CanESM2_historical+rcp26_r1i1p1_19500101-21001231 {
dimensions:
	lon = 1068 ;
	lat = 510 ;
	time = 55115 ;
 variables:
         ...
//global attributes
          ...
}
```

But if this code isn't actually tackling that issue, obviously this isn't an issue. :)

rod-glover:
The demo has been fixed (wrong config, argh). @corviday , you were right that what you were seeing was due to that problem. So have at it!

Dockerfile
```diff
@@ -2,7 +2,7 @@
 # Dockerfile to run the PCIC data portal   #
 ############################################
 
-FROM ubuntu:17.10
+FROM ubuntu:18.04
```
rod-glover: Is this OK?

pdp/static/js/__test__/date-filter-tests.js
```diff
@@ -0,0 +1,374 @@
...
+function resetAllDataServices() {
+    dataServices.getMetadata.reset();
+    dataServices.getCatalog.reset();
+    dataServices.getNCWMSLayerCapabilities.reset();
+    dataServices.getNcwmsLayerDDS.reset();
+    dataServices.getNcwmsLayerDAS.reset();
+}
+
+function dateFilterTests(
+    app, config
```
rod-glover: Document this beast, especially `config`.

corviday:
The demo looks good. I appreciate all the extra work it took to compensate for the missing calendar-style datepicker interface: displaying the start and end dates, error messages when people enter invalid dates, etc. I wasn't able to trip it up despite some efforts to the contrary; all the events seem to fire in the right order.

rod-glover:
@corviday , I checked that the download link time ranges were correct for a few cases. But since this is basically the core of the issue, would you mind kicking the tires a few times on this if you haven't already?

corviday:
I have played around with the time ranges a fair bit; they look good to me.

pdp/static/js/__test__/app-test-helpers.js
```diff
@@ -0,0 +1,18 @@
+function initializeGlobals() {
+    window.pdp = {
+        VERSION: '0.1',
+        app_root: 'https://data.pacificclimate.org/portal',
+        data_root: 'https://data.pacificclimate.org/data',
```
corviday: So does this mean tests will be run using the live backend, and not the development backend? I don't think that's a huge issue that needs to be corrected now, but maybe make a note of that fact so that a hypothetical person changing the back end and front end in tandem would understand why they're getting anomalous test results?

pdp/static/js/__test__/pdp_raster_map.test.js
```diff
@@ -0,0 +1,106 @@
...
+                each(calNames).it('to %s: transfers', function (toName) {
+                    expect(transferDate(date, cfTimeSystem[toName], fallbackDate[toName]))
+                    .toEqual(cfDatetime[toName](year, month, day));
+                })
+            });
+        });
+    });
+
+
+    describe('tricky dates', function () {
```
corviday: 👍

corviday:
Overall this seems pretty solid to me! I want to look over the test code a bit more in the morning, but everything seems to be in good shape.

pdp/static/js/__test__/app-test-helpers.js
rod-glover: Augh, you're right! Tests should not depend on live data, for the reason you give! More mocking! Augh!!

rod-glover: Actually, I was right. These just match the mock data, which looks like but is not real data. Yay. Fixing to make more apparent.

rod-glover:
@corviday , ready for your next review. Thanks for your diligence!

rod-glover:
There's an intermittent error that looks like it's due a problem in the Google Chrome distro, maybe they didn't update their hash sums, or they are doing some kind of CD and the hash sums are lagging or cached or something (???).

Anyway it's not a fault in our code. If it persists, we should probably file an issue with TravisCI.

```
Reading package lists...
E: Failed to fetch http://dl.google.com/linux/chrome/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo apt-get update -q" failed and exited with 100 during .
```

corviday:
I ran the frontend tests and got this error:

```
    console.error node_modules/jsdom/lib/jsdom/virtual-console.js:29
      Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)
          at module.exports (/home/lzeman/Code/pdp-test/pdp/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
          at HTMLCanvasElementImpl.getContext (/home/lzeman/Code/pdp-test/pdp/node_modules/jsdom/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js:42:5)
          at HTMLCanvasElement.getContext (/home/lzeman/Code/pdp-test/pdp/node_modules/jsdom/lib/jsdom/living/generated/HTMLCanvasElement.js:50:45)
          at eval (eval at <anonymous> (/home/lzeman/Code/pdp-test/pdp/pdp/static/js/__test__/globals-helpers.js:68:13), <anonymous>:4082:39)
          at eval (eval at <anonymous> (/home/lzeman/Code/pdp-test/pdp/pdp/static/js/__test__/globals-helpers.js:68:13), <anonymous>:4083:3)
          at eval (<anonymous>)
          at globalEval (/home/lzeman/Code/pdp-test/pdp/pdp/static/js/__test__/globals-helpers.js:68:13)
          at Array.forEach (<anonymous>)
          at Object.forEach [as importGlobals] (/home/lzeman/Code/pdp-test/pdp/pdp/static/js/__test__/globals-helpers.js:58:11)
          at Object.importGlobals (/home/lzeman/Code/pdp-test/pdp/pdp/static/js/__test__/canada_ex_app.test.js:5:30) undefined
```

Is that expected and generally known to be harmless?

pdp/static/js/__test__/pdp_controls.test.js
```diff
@@ -0,0 +1,71 @@
...
+        afterEach(function() {
+            dlLink = null;
+        });
+
+        it('has the supplied initial time range', function() {
+            expect(dlLink.trange).toBe('t0:t1');
+        });
+
+        it('generates the expected download URL', function() {
+           expect(dlLink.getUrl()).toBe('.nc?tasmax[t0:t1][y0:y1][x0:x1]&')
```
corviday: Just checking, but this test generates a download URL with an undefined filename?

pdp/static/js/pdp_controls.js
```diff
@@ -115,13 +224,38 @@ var getRasterControls = function (ensemble_name) {
     return div;
 };
 
-var getRasterDownloadOptions = function (include_dates_selection) {
+
+function cfDateTimeFor(cfTimeSystem, specifier) {
+    if (_.isString(specifier)) {
```
corviday: This function could use a docstring.

pdp/static/js/pdp_raster_map.js
```diff
@@ -1,223 +1,115 @@
...
+    // We associate, using jQuery's `.data()` method, a date value
+    // to the start and end datepicker elements. This value encodes the meaning
+    // of the date entered in each element as a `CfDatetime`, which carries
+    // all information about the CF time system (units, start date,
+    // calendar, index count) as well as the value of the date proper.
+    //
+    // We attempt to transfer the existing dates to the next time system.
+    //
+    // Note: we don't have yearRange any more.
+    // TODO: Find out what yearRange was for and compensate.
```
corviday: I think `yearRange` was a UI parameter that told the calendar-style datepicker which years should be present in its dropdown menu. You've providing that functionality via checking date validity, so I don't think we need to worry about any missing functionality from `yearRange`.

PR #125, opened 2019-05-31
Update Daily Gridded Meteorological Datasets app

rod-glover:
This PR updates `gridded_observations_app`, which provides the Daily Gridded Meteorological Datasets portal. 

This PR adds testing for `gridded_observations_app`, based on the infrastructure established for testing `canada_ex_app`. Both the changes to the app and the new test code are quite simple, with the following observations:

- Data services have to be mocked differently for different apps. Therefore the mocks have been placed in a subdirectory of `__mocks__`, each one named for the app it serves. This departs from the [Jest module "manual mocks" ](https://jestjs.io/docs/en/manual-mocks) arrangement, deliberately, since we can no longer use it.


rod-glover:
[Demo on docker.](http://docker-dev02.pcic.uvic.ca:30508/gridded_observations/map/)

PR #128, opened 2019-06-13
Update Gridded Hydrologic Model Output app

rod-glover:
This PR updates `vic_gen1_app`, which provides the Gridded Hydrologic Model Output portal. 

This PR adds testing for `vic_gen1_app`, based on the established test infrastructure.

It also corrects a couple of minor testing problems for other apps.


rod-glover:
[Demo](http://docker-dev01.pcic.uvic.ca:30509/hydro_model_out/map/)

corviday:
The "historical" scenarios only runs to 2006, but when I select one of them, the end date is displayed as 2099-12-31. This error occurs in the live portal too, and doesn't seem to have been introduced by these changes, but it's much more obvious now than it was, since this portal explicitly displays the end date.

rod-glover:
That looks like a data problem. The app hits the "DDS" API endpoint to get the length of the time dimension. 

For a non-historical dataset, it returns:

```
Dataset {
    Int32 time[time = 54787];
} 5var_day_CCSM3_A1B_run1_19500101-20991231%2Enc;
```

For a historical dataset, it returns the same length:

```
Dataset {
    Int32 time[time = 54787];
} 5var_day_BASE_historical_run1_19500101-20061231%2Enc;
```

This goes all the way back to the data file itself:

```
rglover@pcic-2022:/storage/data/climate/hydrology/vic/gen1/output$ ncdump -h 5var_day_BASE_historical_run1_19500101-20061231.nc
netcdf \5var_day_BASE_historical_run1_19500101-20061231 {
dimensions:
	time = 54787 ;
	lon = 215 ;
	lat = 163 ;
```

rod-glover:
@corviday, all good otherwise?

corviday:
Huh, weird. The file is apparently all fill values after 2006. Not an issue with the code, then.
Dec 31 2006:
31-Dec-2006
Jan 1 2007:
1-Jan-2007

Everything else seemed to be working great.

PR #143, opened 2019-10-23
Add Jenkinsfile

nikola-rados:
A Jenkinsfile has been added at the top level directory. This controls the Jenkins pipeline which is activated on each push to this repo (github webhook).  The pipeline runs through the `npm ` and `python` test suites, then a docker image is created under the name `pdp/[branch_name]`.


nikola-rados:
@corviday if you would be able to run this PR through the same test you have done before that would be great!

PR #148, opened 2019-12-03
fixfilter resetting leads to empty date strings

### LINKED ISSUE #147: On "Reset Filters" map tiles disappear
When I am using the [BC station data portal](https://data.pacificclimate.org/portal/pcds/map/) and click the Reset Filters button, all map tiles disappear and show a missing image symbol instead. When I then resupply a filtering date range, the map tiles reappear. Seems that what is shown on the map is contingent on the date including the base layers. The behaviour I would expect is that when I click on the Reset Filters button I get a map that looks like the original when the page first loads. That is zoomed out to BC wide and showing all stations. Instead of killing the date range, Reset Filters would be better off if it extended the date range back out to the fully available date range for data in BC.

faronium:
Patch to fix an issue where filter resetting leads to empty date strings that throw date errors in the UI and cause the WMS to fail.

Fixes #147 

Signed-off-by: Faron Anslow <fanslow@uvic.ca>

### DIFF
--- a/pdp/static/js/crmp_controls.js
+++ b/pdp/static/js/crmp_controls.js
@@ -128,6 +128,11 @@ function getResetButton(map) {
     $(reset_button).bind('click', function () {
         this.form.reset();
         filter_clear(map);
+        // This wipes the date filters so the call to dateChange gets an empty string and
+        //throws ugly errors that eventually leads to erroneous tile loading from WMS.
+        map.filters.date = date_filter('YYYY/MM/DD','YYYY/MM/DD');
+        $('#from-date')[0].value=map.filters.date.filters[0].value
+        $('#to-date')[0].value=map.filters.date.filters[1].value
         dateChange(map);
         CRMPFilterChange(map);
         return false;
--- a/pdp/static/js/crmp_filters.js
+++ b/pdp/static/js/crmp_filters.js
@@ -15,10 +15,10 @@ function net_filter(net_name) {
 
 function date_filter(sdate, edate) {
     var d = new Date();
-    if (edate === 'YYYY/MM/DD') {
+    if (edate === 'YYYY/MM/DD' || edate === '') { //just in case still getting an empty string
         edate = d.getFullYear() + '/' + (d.getMonth() + 1) + '/' + d.getDate(); //today
     }
-    if (sdate === 'YYYY/MM/DD') {
+    if (sdate === 'YYYY/MM/DD' || sdate === '') { //here as well
         sdate = '1870/01/01'; // ~beginning of our data (unless we find a magic data set)
     }
     // http://stackoverflow.com/questions/325933/determine-whether-two-date-ranges-overlap
@@ -27,12 +27,12 @@ function date_filter(sdate, edate) {
         filters: [
             new OpenLayers.Filter.Comparison({
                 type: ">=",
-                property: "max_obs_time",
+                property: "min_obs_time", //this was wrong in previous version, the sdate in the min obs time.
                 value: sdate
             }),
             new OpenLayers.Filter.Comparison({
                 type: "<=",
-                property: "min_obs_time",
+                property: "max_obs_time", //this was wrong in previous version, the edate in the max obs time.
                 value: edate
             })
         ]


jameshiebert:
Thanks for taking a shot at this @faronium ! I'll let Rod handle the review as our JavaScript guru :)

pdp/static/js/crmp_filters.js
```diff
@@ -27,12 +27,12 @@ function date_filter(sdate, edate) {
         filters: [
             new OpenLayers.Filter.Comparison({
                 type: ">=",
-                property: "max_obs_time",
+                property: "min_obs_time", //this was wrong in previous version, the sdate in the min obs time.
```
rod-glover: Lose these changes. 

The original code only *looks* wrong. It's actually right, given that the test is whether the date ranges *overlap*. See the SO [question](http://stackoverflow.com/questions/325933/determine-whether-two-date-ranges-overlap) mentioned in the code comment just above.

pdp/static/js/crmp_filters.js
```diff
@@ -27,12 +27,12 @@ function date_filter(sdate, edate) {
...
                 value: sdate
             }),
             new OpenLayers.Filter.Comparison({
                 type: "<=",
-                property: "min_obs_time",
+                property: "max_obs_time", //this was wrong in previous version, the edate in the max obs time.
```
rod-glover: Ditto

pdp/static/js/crmp_controls.js
```diff
@@ -128,6 +128,11 @@ function getResetButton(map) {
     $(reset_button).bind('click', function () {
         this.form.reset();
         filter_clear(map);
+        // This wipes the date filters so the call to dateChange gets an empty string and
+        //throws ugly errors that eventually leads to erroneous tile loading from WMS.
+        map.filters.date = date_filter('YYYY/MM/DD','YYYY/MM/DD');
+        $('#from-date')[0].value=map.filters.date.filters[0].value
+        $('#to-date')[0].value=map.filters.date.filters[1].value
```
rod-glover: Let's put this into `filter_clear()` instead of here. `filter_clear` is only called in one place, here, but I think this logic more rightly belongs inside it, not outside it. That function should be revised to look like:
```
function filter_clear(map) {
    var date = date_filter('YYYY/MM/DD','YYYY/MM/DD');
    map.filters = {
        values: map.filters.values,
        date: date
    };
    $('#from-date')[0].value = date.filters[0].value;
    $('#to-date')[0].value = date.filters[1].value;
    map.getLayersByName("Polygon selection")[0].removeAllFeatures();
}
```

I don't particularly like mixing the setting of `map.filters` with the manipulation of the DOM (`$(...)`), but the PDP code mixes concerns all the time and I think this is the right place for it.

pdp/static/js/crmp_filters.js
```diff
@@ -15,10 +15,10 @@ function net_filter(net_name) {
 
 function date_filter(sdate, edate) {
     var d = new Date();
-    if (edate === 'YYYY/MM/DD') {
+    if (edate === 'YYYY/MM/DD' || edate === '') { //just in case still getting an empty string
```
rod-glover: Very likely unnecessary, but hey, belt and suspenders.

rod-glover:
@faronium , review completed. Let me know if you'd like me to help in any other way here.

pdp/static/js/crmp_controls.js
faronium: Thanks for looking through this. I don't know much about javascript and even less about how the DOM works, but, I couldn't come up with another way, in my hacking around, to force populate the date range text boxes so that the page looked like it does at start.

That brings up a potential nuclear and inelegant solution which would be to force a reload of the page when the reset filters button is pressed...

rod-glover: I think your solution is likely fine. My only quibble was with how the code is organized. 

Did you run this changed code to check it out?

faronium:
> @faronium , review completed. Let me know if you'd like me to help in any other way here.

Thanks Rod and James. Nope, no other help. If you feel like re-deploying the crmp/pcds portal with this fix, that would be great. But, I'll leave it to you to decide if you want to merge this into what's currently running. My preference would be to do so given that I find it a bit annoying myself when I use the portal, let along how our users may feel.

pdp/static/js/crmp_controls.js
faronium: No, I did all my debugging initially through the running application on firefox. Inserted breakpoints in the code then forced the changes to occur in the console. I don't think it would be wise for me to try to get this running on my laptop as I know I would spend hours getting it all configured... :grimacing:

faronium: running your change in the js console in firefox gives the desired result. I put a breakpoint at crmp_filters.js:152, hit "Reset Filters" on the crmp app and when the execution stops at the break, paste in your changes and then resume the execution and behaviour is as desired.

So, not a proper test, but seems to be strong support that it is working.

PR #149, opened 2019-12-06
Jenkins Pipeline Update

nikola-rados:
This update applies the new features found in the [`climate-explorer-frontend Jenkinsfile`](https://github.com/pacificclimate/climate-explorer-frontend/blob/master/Jenkinsfile).

Features added:
- tag tracking
- reduced usage of security scan
- more consistent image naming

These features are found in two other projects and will eventually be factored out.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,7 +42,7 @@ def run_tests() {
 def build_image(image_name) {
     def image
     withDockerServer([uri: PCIC_DOCKER]) {
-        image = docker.build(image_name)
+        image = docker.build(image_name, '--pull .')
     }
 
     return image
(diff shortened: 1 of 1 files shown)


nikola-rados:
The functionality added was to collect and process the git tags.  These changes do not just apply to `master`.  If the branch is `master` and is tagged on git, we want to push the git tag(s) along with the `latest` tag to docker hub.  If the branch is not `master` we use the branch name as the tag.

nikola-rados:
Rather than open _another_ PR for _another_ `Jenkinsfile` update, I thought I would extend this one.  The changes in the most recent commit maintain the same functionality added by this PR, but do so using the shared library.

nikola-rados:
@corviday that was not intended.  Did you produce that image?

nikola-rados:
I re-ran the pipeline (with a small change that wasn't relevant to image naming) and it produced the expected output.  Would you be able to tell me how you produced the result you got?

corviday:
Hm, I might have branched off the wrong place for my tests. I'll try again.

PR #150, opened 2019-12-06
Fix date reset

### LINKED ISSUE #147: On "Reset Filters" map tiles disappear

rod-glover:
Resolves #147 

[Demo](http://docker-dev01.pcic.uvic.ca:30508/pcds/map/)

### DIFF
--- a/pdp/static/js/__test__/crmp_app.test.js
+++ b/pdp/static/js/__test__/crmp_app.test.js
@@ -78,6 +78,40 @@ dateFilterTests(crmp_app, {
 });
 
 
+describe('CRMP app', function() {
+    beforeEach(function () {
+        // Reset the DOM (jsdom)
+        document.body.innerHTML = htmlTemplate();
+        mockHelpers.resetAll(dataServices);
+        mockHelpers.resolveAllWithDefault(dataServices);
+        crmp_app();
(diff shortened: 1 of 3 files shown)


corviday:
Can you put the demo back up? I am hoping to release a new PDP in the next day or two, and I'd love to get this fix into it. Code looks good, but I wanted to poke at the demo as well, and I think the docker-dev01 upgrade last week took it down.

rod-glover:
[Demo](http://docker-dev01.pcic.uvic.ca:30508/pcds/map/) up

faronium:
Looks great Rod. Thanks for getting to this. Looking forward to the roll-out.
--Faron

PR #156, opened 2020-01-21
Add new hydrology data to the portal

corviday:
This PR updates the `hydro_model_out` portal, which serves the VIC generation 1 data, to serve the new VIC generation 2 data as well.

The two VIC datasets are presented very similarly. They have seperate backends (`vic_gen1.py` and `vic_gen2.py`) but share the same front-end code (`vic_app.js`, `vic_controls.js`, and `vic_map.js`). Each portal links to the other one. The front end determines whether it is serving gen1 or gen2 by the URL it was called with, which is a little ugly, but worth it to not have to copy and paste a second portal. (The same arrangement - separate backends, shared frontend - is also used by the portal that serves the BCCAQ and BCCAQ2 data.)

This PR also contains a couple small changes that affect code outside this portal: 
* on all portals, items in the dataset selection menu tree will now be alphabetized at each level
* notes on how to set up jest mocking for tests for the portals that display multiple datasets (this was very confusing)
* in the map legend, datasets whose units are `1` will have their units displayed as `fraction`. So far, the only data with units of `1` is glacier cover data served by the new portal

A demo is available [here](http://docker-dev02.pcic.uvic.ca:30555/hydro_model_out/map/).


jameshiebert:
Everything looks great to me. Good work!

> The front end determines whether it is serving gen1 or gen2 by the URL it was called with, which is a little ugly

Seems fine to me, unless @rod-glover has better ideas.

doc/source/raster.rst
```diff
@@ -200,7 +200,7 @@ You can retrieve the catalog with your favorite programming languages as well. F
   > json_data[[1]]
   [1] "https://data.pacificclimate.org/data/downscaled_gcms/pr+tasmax+tasmin_day_BCCAQ+ANUSPLIN300+GFDL-ESM2G_historical+rcp45_r1i1p1_19500101-21001231.nc"
 
-At present, there are five pages for which one can retrieve catalogs: ``bc_prism``, ``downscaled_gcms``, ``downscaled_gcm_extremes``, ``gridded_observations`` and ``hydro_model_out``.
+At present, there are seven pages for which one can retrieve catalogs: ``bc_prism``, ``downscaled_gcms``, ``downscaled_gcmws_archive``, ``downscaled_gcm_extremes``, ``gridded_observations``, ``hydro_model_archive``, and ``hydro_model_out``.
```
rod-glover: gcmws? Probably right, but I've never seen this acronym before.

pdp/static/js/__test__/README.md
```diff
@@ -0,0 +1,57 @@
...
+#### getNCWMSLayerCapabilities
+A representation of ncWMS's `GetCapabilities` query, a very long xml document.
+Needs to include, at minimum, the unique ID of the dataset in question, and the
+default timestamp, which can be found in the portal's initialization javascript, 
+in their usual places in the `<LAYER>` element.
+
+The xml includes a long list of available palettes; all but the default one
+specified in the portal's map initial code are skippable and do not need
+to be mocked (`boxfill/ferret`for most variables; `boxfill/blueheat` for some
+precipitation datasets).
```
rod-glover: Thank you for documenting this! Nice and clear!

pdp/static/js/pdp_controls.js
```diff
@@ -171,7 +171,23 @@ function processDateRangeInput($date, fallbackFlag, $error) {
...
+            leafNameMapping[a] : a).toLowerCase();
+        const bdn = (leafNameMapping && b in leafNameMapping ?
+            leafNameMapping[b] : b).toLowerCase();
+        if(adn < bdn) {
+            return -1;
+        }
+        if(adn > bdn) {
+            return 1;
+        }
+        return 0;
```
rod-glover: You can replace the whole set of comparisons and returns with [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

```
return adn.localeCompare(bdn);
```

pdp/static/js/pdp_map.js
```diff
@@ -47,8 +55,8 @@ function BC3005_map_options() {
     return options;
 }
 
-function BC3005_map_options_vic() {
-    var bounds = getBC3005Bounds_vic(),
+function BC3005_map_options_vic(archive_portal) {
+    var bounds = getBC3005Bounds_vic(archive_portal),
```
rod-glover: It appears that the argument `archive_portal` isn't used in this function. Is it needed for some non-obvious reason?

pdp/static/js/vic_map.js
```diff
@@ -1,14 +1,30 @@
...
+	const vic_gen2_init = {
+		dataset: "BASEFLOW_day_VICGL_ACCESS1-0_rcp85_r1i1p1_19450101-20991231_columbia",
+		variable: "BASEFLOW",
+		time: "2000-01-01T00:00:00Z"
+	};
+	const init = archive_portal ? vic_gen1_init : vic_gen2_init;
+
     // Map Config
-    options = BC3005_map_options();
+    options = BC3005_map_options_vic(archive_portal);
```
rod-glover: `BC3005_map_options_vic` doesn't use its argument (at least so it seems). If not, remove it here too?

pdp/static/js/vic_map.js
```diff
@@ -42,7 +53,7 @@ function init_vic_map() {
         pdp.ncwms_url,
         params,
         {
-            maxExtent: getBC3005Bounds_vic(),
+            maxExtent: getBC3005Bounds_vic(archive_portal),
```
rod-glover: Ditto

pdp/static/js/vic_map.js
```diff
@@ -60,7 +71,7 @@ function init_vic_map() {
     );
 
     document.getElementById("pdp-map").appendChild(getOpacitySlider(ncwms));
-    map.zoomToExtent(new OpenLayers.Bounds(611014.125, 251336.4375, 2070975.0625, 1737664.5625), true);
+    map.zoomToExtent(getBC3005Bounds_vic(archive_portal));
```
rod-glover: Ditto

pdp/static/js/pdp_controls.js
corviday: That's great. when I was typing it, I was sure there must be a better way.

pdp/static/js/pdp_map.js
corviday: It's passed to `getBC3005Bounds_vic()` to determine the map bounds, a fact made unclear due to me naming the argument to `getBC3005Bounds_vic()`  something unrelated-seeming, instead of `archive_portal`, which is what this predicate is called everywhere else in the codebase. (https://github.com/pacificclimate/pdp/pull/156/files/a591efa2575be657d05c5b368181dfe1edb9d253#diff-708844e00533945241e16a8ef970f907R20). Thanks for catching that, I'll clear that up.

rod-glover:
LGTM!

PR #168, opened 2020-04-02
Update station data

corviday:
This PR Is to solicit @rod-glover 's help with some trouble I am having with OpenLayers.

**What I am trying to do**: change the map window of the Hydro_station portal to extend farther north. 
**How I am trying to do it**: increase the 4th numerical value on line 79 of pdp_map.js , which is used as bounds for the map
**What happens**: Instead of merely extending north, the entire view moves north, cutting off the southern data.

What should I do differently?


corviday:
Possibly helpful notes:

* this branch is running a demo here: http://docker-dev02.pcic.uvic.ca:30555/hydro_stn_cmip5/map/
* the portal displays two different sets of hydrology stations. The "archive" dataset is displaying fine; the non-archive dataset is the one I am. having trouble getting the map bounds set for

rod-glover:
It's possible you are running afoul of or should instead be setting the `maxExtent` option for [`OpenLayers.Map`](http://dev.openlayers.org/releases/OpenLayers-2.13.1/doc/apidocs/files/OpenLayers/Map-js.html#OpenLayers.Map.OpenLayers.Map).

> maxExtent
{<OpenLayers.Bounds>|Array} If provided as an array, the array should consist of four values (left, bottom, right, top).  The maximum extent for the map.  Default depends on projection; if this is one of those defined in OpenLayers.Projection.defaults (EPSG:4326 or web mercator), maxExtent will be set to the value defined there; else, defaults to null.  To restrict user panning and zooming of the map, use restrictedExtent instead.  The value for maxExtent will change calculations for tile URLs.

> restrictedExtent
{<OpenLayers.Bounds>|Array} If provided as an array, the array should consist of four values (left, bottom, right, top).  Limit map navigation to this extent where possible.  If a non-null restrictedExtent is set, panning will be restricted to the given bounds.  In addition, zooming to a resolution that displays more than the restricted extent will center the map on the restricted extent.  If you wish to limit the zoom level or resolution, use maxResolution.

I particularly call out the sentence "Limit map navigation to this extent where possible." This suggests to me that `restrictedExtent` is subordinate to `maxExtent`, and its purpose is to limit, not extend or define, the map extent. I think the latter is the effect you are looking for. 

Consider where `maxExtent` is set in `pdp_map.js`: https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/pdp_map.js#L163 

This may be your issue.

rod-glover:
There's also another place where `maxExtent` is set, but I think it doesn't apply to your map: https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/pdp_map.js#L117

corviday:
Thank you for your help. I think, following your logic, I have turned up the issue. it seems like OpenLayers has discrete zoom levels, one of which is just _slightly_ too small to fit all the data stations, which is what led to my feeling of being able to move the view around by changing the `Bounds` values, but not actually expand it enough to cover all the stations.

I was using the `zoomToExtent()` function with `close`=`true`, which told OpenLayers to use the _closest_ map scale to what I requested, so it kept using the slightly-too-small zoom level. Setting `close` to `false` results in a map that seems very zoomed out, but does fit all the data, and people can zoom into the area they want, I guess.

PR #169, opened 2020-04-03
Add new routed streamflow data

### LINKED ISSUE #163: Add new data to hydro station portal
Hydrology has some CMIP5-based data to add to the hydro station portal. I need to check, but I think I remember that, like the hydro model output portal, they want the previous set of data to stay accessible as well.

corviday:
This PR adds new CMIP5 streamflow data to the Routed Streamflow Data portal. The old data is still present as an "archive" portal, accessible via a link from the new portal. The up to date portal and the archive portal have separate python backends but share a javascript frontend, as the UI considerations are the same for both datasets. The PDP determines whether it is serving archive or normal data by checking for the presence of the word "archive" in the URL it is being accessed with, which is straightforward but a little ugly.

Resolves #163


corviday:
Demo: http://docker-dev02.pcic.uvic.ca:30555/hydro_stn_cmip5/map/

pdp/static/js/pdp_map.js
```diff
@@ -55,6 +55,33 @@ function BC3005_map_options() {
...
+}
+
+
+function getBC3005Bounds_stn(archive_portal) {
+	if(archive_portal) {
+		// archive portal's dataset cover watersheds in BC only
+		return new OpenLayers.Bounds(611014.125, 251336.4375, 2070975.0625, 1737664.5625);
+	}
+	else {
+		// new dataset includes entire Columbia watershed south to NV, USA
```
jameshiebert, line 78: I had no idea that the Columbia made it into Nevada. Craziness.

jameshiebert:
I've read through the code and it all looks straightforward to me. I don't have any substantive feedback or constructive critique, so I'd be fine if you go ahead and merge it.

PR #180, opened 2020-08-06
Add actions

### LINKED ISSUE #179: Add actions
Move the project's CI from `jenkins` to `actions`.

nikola-rados:
## Changes
- Add `docker` publishing workflow
- Add `anchore` scan workflow
- Add `node` test workflow
- Add `pypi` publishing workflow
- Add `python` test workflow
- Update `.gitignore`
- Remove `travis`
- Heavily reduce `Jenkins` responsibilities
- Mark `pytest` tests

## Notes
This is a big step towards replacing `Jenkins` in this repo. The only thing `Jenkins` remains responsible for is running tests that require a database connection.

Resolves #179.


pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+markers =
+    slow: marks tests that are slow (deselect with '-m "not slow"')
+    crmpdb: marks tests that use a PCIC database (deselect with '-m "not crmpdb"')
```
corviday, line 4: This is a confusing name for this marker. the PDP has two databases, `crmpdb` (weather station data) and `pcic_meta` (netCDF data). If I was writing a new test using the database to do some netCDF stuff, I probably wouldn't think to mark it `crmpdb`, since the test has nothing to do with `crmpdb`. Could you name it something that refers to both databases like like `database` or `needsdb` or something?

corviday: Or I guess you could document it somehow so developers knew.

corviday: Hm, I think probably all the tests that use the `pcic_meta` database are already marked `bulk_data`, which means they require datafiles on `/storage`. So perhaps instead of extending the `crmpdb` tag to a bunch of tests that don't use that database, you could just exclude the tests tagged `crmpdb` _and_ the tests tagged `bulk_data` when you want to exclude all tests that use a live PCIC database.

nikola-rados: >  you could just exclude the tests tagged crmpdb and the tests tagged bulk_data when you want to exclude all tests that use a live PCIC database.

I think this is the best way to go about it.

PR #186, opened 2020-11-12
Add dockerignore

### LINKED ISSUE #185: Add dockerignore

nikola-rados:
There is a small issue with `get_commitish()` when we apply a `.dockerignore` file. Since we remove the `.git` from the build context there is no git info for the method to pull. I have put a bandaid on it for the moment, but I would like to discuss other options.

Resolves #185.


rod-glover:
I see what you're doing, and it seems reasonable enough, but maybe unnecessary?

What are your concerns with the `gitpython` solution? My own would be:
- [Leakage of System Resources](https://gitpython.readthedocs.io/en/stable/intro.html#leakage-of-system-resources): "GitPython is not suited for long-running processes (like daemons) as it tends to leak system resources." PDP is definitely long-running.

But a bigger question is:
- The `docker-publish` action pulls the repo, so there is a `.git` directory in the larger context. And the original way this was done was that `git_commitish` was a shell script (so with access to `.git`, right?) whose value is passed in to the docker build as a build arg. You set something equivalent up in [Plan2Adapt](https://github.com/pacificclimate/plan2adapt-v2/blob/master/.github/workflows/docker-publish.yml). I see that it's a trifle tricky, but obvious enough. Is that not an option here?

Asides: 
- You don't specify `gitpython` as a dependency in `setup.py.` Was that intentional? Perhaps it would be a bit circular, or at least cart-before-horse there?
- I'm going to assume that the Jenkins fails don't matter.

rod-glover:
@nikola-rados : Not a big deal, but I have an interest in PDP: What do you make of the alternative (maybe cleaner) method that I noted is used in P2A and elsewehere?

rod-glover:
@nikola-rados , it looks like this PR was never merged. I think whatever else is done we still need a `.dockerignore`, but we don't have one in the master branch.

nikola-rados:
Oh sorry, I lost your comments in my ocean of emails. 

I think the alternative is the way to go. I'll restore the branch and see if I can get things working.

PR #188, opened 2020-11-26
Modify ncWMS requests for ncWMS 2.x

### LINKED ISSUE #187: Modify ncWMS requests for ncWMS 2.x
PDP presently makes ncWMS 1.x requests. Our ncWMS service is now running ncWMS 2.x, and the following changes to requests need to be made:

(See PR for list of fixes needed).

Note that the ncWMS-mm-rproxy translation service enables PDP to use unique id's and remain ignorant of dynamic datasets.

rod-glover:
Resolves #187

[Demo](http://docker-dev02.pcic.uvic.ca:30555/downscaled_gcms/map/):
- This version of PDP making requests to ...
- The translation service, which is using the `pcic_meta` database, and makes requests to ...
- The ncWMS 2.5.1 dynamic datasets instance, which has been updated to mount the PDP (`pcic_meta`) datasets.

Testing:
- [x] [BC Station Data - PCDS](http://docker-dev02.pcic.uvic.ca:30555/pcds/map/): does not use ncWMS
- [x] [High-Resolution PRISM Climatology](http://docker-dev02.pcic.uvic.ca:30555/bc_prism/map/)
  - presently fails because of dataset file issues (non-montonic time dimension)
  - [x] Get input from Faron re. datasets and UI
  - [x] Update datasets and reindex them
  - [x] Implement UI changes
- [x] [Statistically Downscaled GCM Scenarios](http://docker-dev02.pcic.uvic.ca:30555/downscaled_gcms/map/)
  - map layers work
  - fails on GetLegendGraphic - needs analysis. 
    - Problem: `NUMCOLORBANDS=254`, > 249 
    - Fixed
- [x] [Daily Gridded Meteorological Datasets](http://docker-dev02.pcic.uvic.ca:30555/gridded_observations/map/)
  - map layers work
  - fails on GetLegendGraphic - fixed
  - palette = `default` is different
    - `x-Rainbow` appears to be identical ncWMS 2 predefined palettes: fixed
- [x] [Gridded Hydrologic Model Output](http://docker-dev02.pcic.uvic.ca:30555/hydro_model_out/map/)
  - [x] needs access to `/storage/data/projects/hydrology/dataportal/CMIP5/VICGL`
  - [x] file not found 
    - confirmed mount of `/storage/data/projects/hydrology/dataportal`
    - confirmed file exists
    - executing `ls` in container shows (a) file exists, (b) permission denied.
    - VICGL directory lacks `x` permission for all users. Changed permission. Files now accessible.
  - [x] GetLegendGraphic and GetMap requests `undefined` palette - error for ncWMS 2 (ncWMS 1 evidently used `default`)
  - [x] GetMap requests `SRS=EPSG:3005` - error:
    ```
        The CRS EPSG:3005 is not supported by this server
        Cause:
            No &#8216;CoordinateReferenceSystem&#8217; object found for code &#8220;3005&#8221;. However only a subset of the EPSG geodetic dataset has been queried. See http://sis.apache.org/epsg.html for instruction about how to install the full EPSG database.

    ```
- [x] [Modelled Streamflow Data](http://docker-dev02.pcic.uvic.ca:30555/hydro_stn_cmip5/map/): does not use ncWMS

Fixes required to PDP:
 
- [x] `STYLES` parameter: `boxfill` renamed `default`
- [x] `NUMCOLORBANDS` parameter: Values limited to 1 ... 249: Use 249 (not 254)
- [x] Use `x-Rainbow` palette instead of `default`
- [x] Replace `undefined` palette with `x-Rainbow` (Gridded Hydrologic Model Output)
- [x] Improve configuration for BC PRISM portal to handle new datasets.

Fixes required externally:

- [x] Add custom palettes to ncWMS 2 to support absent ncWMS 1 palettes used in PDP. 
- [x] Load full EPSG geodetic dataset into ncWMS 2 instances.
- [x] Split up PRISM datasets into single-resolution files.


rod-glover:
# Questions re. PRISM portal

Below are side-by-side screenshots of current (production) PDP and demo of new PDP. 

TL;DR: The climatology layers look quite different in the new, although the monthly timeseries are unchanged. I suspect a data issue, but am not certain. Thoughts?

## Monthly timeseries

The monthly timeseries displays are unchanged (this is good):

Monthly timeseries 1950-2007 Precip  - current (L), new (R)
image

Monthly timeseries 1950-2007 Tmax  - current (L), new (R)
image

Monthly timeseries 1950-2007 Tmax  - current (L), new (R)
image

## Climatologies

The climatologies look quite different. I suspect some kind of data problem, since the scales are identical and the monthly timeseries look identical.

### 1970-2000

1970-2000 Precip - current (L), new (R)
image

1970-2000 Tmax - current (L), new (R)
image

1970-2000 Tmin - current (L), new (R)
image

### 1981-2010

1981-2010 Precip - current (L), new (R)
image

1981-2010 Tmax - current (L), new (R)
image

1981-2010 Tmin - current (L), new (R)
image

rod-glover:
# Discrepancies in Gridded Hydrologic Model Output

Why are these different? The scales are the same, the layer unique_id's are the same. Metadata is the same.

One difference is that ncWMS 2.x doesn't permit more NUMCOLORBANDS > 249; ncWMS 1.x permitted (and we used) 254. Hard to image that would make any difference.

Icky question: Is ncWMS rendering these datasets differently for some reason???

A few examples:

ACCESS1-0 Baseflow
image

CanESM2 Baseflow
image

CanESM2 Precipitation
image

rod-glover:
# No more Modelled Streamflow Data portal?

Prod currently gives 404: https://data.pacificclimate.org/portal/hydro_stn/map/

jameshiebert:
Re: "No more Modelled Streamflow Data portal":

The links from our [website]() show:
https://data.pacificclimate.org/portal/hydro_stn_cmip5/map/ for CMIP5
and
https://data.pacificclimate.org/portal/hydro_stn_archive/map/ for CMIP3

jameshiebert:
Re: "Discrepancies in Gridded Hydrologic Model Output"

I'm comfortable with nominal changes in the visualized representation of the datasets. The goal of the PDP is to *deliver* data, not necessarily to *visualize* data. The maps that are shown are essentially an selected arbitrary single day within a 55,000 time step simulation. So an "accurate" visual representation isn't really meaningful.

tl;dr it's understandable for the plots to change a bit. Just be prepared to explain it if questions come up.

corviday:
Response to [this comment](https://github.com/pacificclimate/pdp/pull/188#issuecomment-749830681). 

**Monthly Timeseries:** 

No discrepencies, this files are unchanged.

**PR Climatologies:** 

Previously, these files contained twelve monthly values and a _much larger_ annual value, with the annual value determining the top of the scale. Presumably the annual value reflected cumulative precipitation across all months. Here's a graph of one of the original, annual-is-cumulative datasets, note the much higher final (annual) value, the one that goes upwards and backwards across the graph: 

image

It's the annual value that pegs the top end of the scale in the old dataset, and presumably the annual value that is displayed on the portal page (it says "June 1996" which could be either the annual value or the monthly value for June). Since the values in the new monthly-only dataset are much lower, I would expect to see something similar to what you show here - values clustered at the low end of the scale - if you use the same colour scale as the monthly-plus-annual dataset with the monthly-only dataset.

Alas, we probably need separate colour scales for the new monthly and annual datasets, ugh.

**Temperature Climatologies**
I'm not sure what's going on here yet, I will examine the old files. I suspect it may be similar, like maybe the annual tmin and tmax value represent whole-year minimums and maximums and make the scales weird.

corviday:
**Temperature Climatologies**

For these datasets, it looks like the annual value is the _mean_ of the monthly values, not anything weirder like the minimum or maximum. Here's one of the old monthly+annual tmin datasets:

tminmonthly+annual

Here's one of the old monthly+annual tmax datasets:

tmax-monthly+annual

My suspicion is that in these cases, the data map discrepancy we're seeing is just caused by switching the displayed timestamp from displaying June (annual timestamp - the mean values, centred nicely in the middle of the scale) to June (monthly timestamp - maximum values at top of scale).

rod-glover:
@corviday , your deductions are completely confirmed by the observation that the annual datasets in the new app exactly match the "monthly" datasets in the old one. Thank you!

docker/local-pytest/up-backend.sh
```diff
@@ -0,0 +1,22 @@
+docker run -it \
+  -e "APP_ROOT=http://0.0.0.0:30555" \
+  -e "DATA_ROOT=http://0.0.0.0:30556" \
+  -e "DSN=postgresql://httpd_meta:gh0_3Basl3Mn@db3.pcic.uvic.ca:5432/pcic_meta" \
```
corviday, line 4: We might want to remove these outdated passwords.

README.md
```diff
@@ -158,6 +169,31 @@ Enable or disable Google Analytics reporting
 
 Google Analytics ID
 
+### JavaScript configuration code
```
corviday, line 172: Thanks for documenting the JS config files!

docker/dev-local/README.md
```diff
@@ -0,0 +1,45 @@
+# Local development Docker deployment
+
+## What
```
corviday, line 3: Meat, I look forward to trying this.

docker/dev-local/prism_demo_config.js
```diff
@@ -0,0 +1,109 @@
+// Configuration file BC PRISM client app.
```
corviday, line 1: This makes me nervous. We have two identical files in the repository, one for testing and one for production, it seems like it would be easy for them to get out of step and throw off testing. I don't have any better ideas than the setup here, though. Maybe the best we can do is to add a comment to this test version noting that it needs to be kept up to date with the other copy in the repository.

pdp/portals/bc_prism.py
```diff
@@ -2,44 +2,85 @@
...
-        return "{} {} {}-{}".format("Monthly " if monthly else "",
-                                    "Climatological Averages " if climatology
-                                    else "Timeseries ",
-                                    date_finder.group(1)[0:4],
-                                    date_finder.group(2)[0:4])
 
+class PrismEnsembleLister(EnsembleMemberLister):
+    
     def list_stuff(self, ensemble):
+        """
```
corviday, line 24: Excellent work on both the docstring update and the clearer description of the datasets for the portal menu.

pdp/static/js/prism_demo_map.js
```diff
@@ -3,9 +3,10 @@
 
 "use strict";
 
-function init_prism_map() {
+function init_prism_map(config) {
     var selectionLayer, options, mapControls, selLayerName, panelControls,
-        map, defaults, params, datalayerName, cb, ncwms;
+      map, params, datalayerName, cb, ncwms;
+    console.log("init_prism_map", config)
```
corviday, line 9: accidentally left debug print?

pdp/static/js/prism_demo_map.js
```diff
@@ -69,37 +65,47 @@ function init_prism_map() {
...
-        } else if (varname == 'tmin' ) {
-            this.params.COLORSCALERANGE = '-15,10';
+        // Extract dataset unique id and variable name from layer name.
+        const layer_name_parts = layer_name.split('/');
+        const uniqueId = layer_name_parts[0];
+        const varname = layer_name_parts[1];
+
+        // Extract metadata from the unique id. This is ugly and fragile,
+        // but since PDP doesn't really have a data model, this is how this
+        // information is available to us here.
```
corviday, line 81: yikes (but I agree we probably have to do it this way)

corviday (approved):
Looks good overall. I had some minor quibbles that could be easily ignored.

docker/dev-local/prism_demo_config.js
rod-glover: This replicates the situation in deployment directories, which is what this really is, or is a template for. We could remove this file and  the container mount of it, but I very much wanted to preserve the record of how the configuration file is overridden, specifically the filepath.

corviday: Makes sense.

pdp/static/js/prism_demo_map.js
jameshiebert: All datasets that are published by the PDP are--by definition--also available from the `modelmeta` database. So we really don't have to extract metadata from the `unique_id` (in fact, I've been advocating against doing so for years). It may be locally *expedient*, but it's absolutely not *necessary*.

rod-glover: Not, I think, without a pretty big change to this codebase.

jameshiebert: I definitely agree that it would require a broader change to the codebase. But both the comments in the code and in the PR say to me "the PDP can't do this any way but through the hacking the unique ID", and that's not true. So at the very least, I think that the code comment should point to the preferred solution that *is* available from the existing data model.

corviday: It might be possible to use [the backend's existing DAS call](https://data.pacificclimate.org/data/gridded_observations/PNWNAmet_wind.nc.das) to get this sort of information in the future, since it reports any global metadata in the file. We'd definitely need to update a bunch of the PDP's datasets to do it, though; many of them predate the adoption of our current standards and have incomplete metadata.

Alternately, we might want to do something more like PCEX and add a new API endpoint that relays information from the modelmeta database.

rod-glover: At the moment, I prefer the latter: https://github.com/pacificclimate/pdp/issues/193

rod-glover: @jameshiebert , your point is well taken, particularly re. the comments. "Have to" is of course false; I intended it relative to the amount of effort it seemed reasonable to put into this particular PR. I've updated the code comments.

rod-glover:
Verifying data downloads:

- [x] [BC Station Data - PCDS](http://docker-dev02.pcic.uvic.ca:30555/pcds/map/): does not use ncWMS
- [ ] [High-Resolution PRISM Climatology](http://docker-dev02.pcic.uvic.ca:30555/bc_prism/map/)
- [ ] [Statistically Downscaled GCM Scenarios](http://docker-dev02.pcic.uvic.ca:30555/downscaled_gcms/map/)
- [ ] [Daily Gridded Meteorological Datasets](http://docker-dev02.pcic.uvic.ca:30555/gridded_observations/map/)
- [ ] [Gridded Hydrologic Model Output](http://docker-dev02.pcic.uvic.ca:30555/hydro_model_out/map/)
- [ ] [Modelled Streamflow Data](http://docker-dev02.pcic.uvic.ca:30555/hydro_stn_cmip5/map/): does not use ncWMS

rod-glover:
@jameshiebert , @corviday : All portals appear to be working correctly. I've verified file downloads from each portal, to the extent I can with a moderate effort. With the discussion and resulting modifications above, do you think that this PR is ready to merge?

corviday:
I think so.

rod-glover:
James also agrees that we are ready to merge.

PR #197, opened 2021-01-19
Update local test docker README based on experience.

corviday:
I had an extremely difficult time getting the test docker setup to work, which turned out to be due to my VPN connection interfering with DNS. I have added a couple sentences to the README covering this possibility.


rod-glover (approved):
LGTM, thanks!

PR #203, opened 2021-03-15
Correct broken PRISM tests

jameshiebert:
PR #188 changed some things around to accommodate ncWMS 2. In order to make the changes some of the base data files (including a few that were referenced in the test suite), were split up and rewritten and no longer exist in the production app.

This changes modifies the test code to match the new data.


rod-glover (commented):
This all looks fine. One possible issue:

File `pdp/static/js/__mocks__/data-services/prism_demo_app.js` includes some mock data  with the name `climatology_bounds` still in them. I *think* that may be irrelevant, since there is an attribute `climatology` that names the variable `climatology_bounds`, but you might want to look at it. Lines 441 - 446.

PR #211, opened 2021-04-30
Update installation

nikola-rados:


nikola-rados:
This is no longer valid

PR #217, opened 2021-06-16
Upgrade to modelmeta >= 0.3.0

rod-glover:
This PR adjusts dependencies to make PDP use `modelmeta` >= 0.3.0. This means it can run against the `pcic_meta` database migrated to revision `12f290b63791`. These changes include:
- Upgrading `modelmeta` and `pdp_util` versions in `requirements.txt`
- Removing version specifications for  `modelmeta` and `pdp_util` from `setup.py`. This was causing build problems and is unnecessary.

[Demonstration](https://services.pacificclimate.org/dev/portal/bc_prism/map/)

In addition to the changes needed for the modelmeta upgrade, there are also the following:
- Fix broken Python CI build. Some changes in the underlying support for our existing build broke the tests. The self-hosted base image has been adjusted to `ubuntu-18.04`, and the build steps changed also to compensate the loss of support. Notes:
  - The base image for the CI is the same as the base image for our production and test Docker images, making them all very similar and consistent.
  - Function `get_commitish()` started raising a git exception in the Python CI build. Online advice on fixing this did not work. Instead, we have a fallback to the value "unknown" if this error occurs. It seems only to occur in the CI build.
- Fix broken production and test Dockerfiles, same reason.
- Fix broken Metadata hyperlink in portals. This fix should have been in a separate PR, but ... The fix itself is 4 characters; the tests to catch this type of error are more bulky.
- Apply Black to `setup.py`, which was getting messy.
- And, oops, a PR for Snyk accidentally merged into this branch. Oh well.


pdp/static/js/pdp_controls.js
```diff
@@ -637,7 +637,7 @@ MetadataDownloadLink.prototype = {
         reg = new RegExp(pdp.data_root + '/(.*)/(.*)');
         matches = reg.exec(url);
         //matches[1] is portal url base, matches[2] is dataset, make catalog URL.
-        url = pdp.app_root + "/" + matches[1] + "/catalog/" + matches[2];
+        url = pdp.data_root + "/" + matches[1] + "/catalog/" + matches[2];
```
corviday, line 640: Tiny bugfix! :)

docker/local-pytest/up-backend.sh
```diff
@@ -18,5 +18,6 @@ docker run -it \
   -e "GUNICORN_TIMEOUT=86400" \
   -e "APP_MODULE=pdp.wsgi:backend" \
   -v $(pwd):/codebase \
+  -v /storage/data:/storage/data:ro \
   --name pdp-local-pytest_backend \
-  pdp-local-pytest
\ No newline at end of file
+  pcic/pdp-local-pytest
```
corviday, line 23: This and `down-backend` seem like useful commands, but not documented anywhere. Could you please mention them in documentation somewhere? Their use is fairly obvious, but the PDP test setup is complicated enough that I am in favour of documenting everything.

pdp/static/js/__test__/pdp_controls.test.js
```diff
@@ -4,69 +4,148 @@ require('./globals-helpers').importGlobals([
...
+                  expect($('#download-metadata').attr('href'))
+                    .toBe(metadata_url)
+              }
+            );
+        });
     });
-});
\ No newline at end of file
+})
+
```
corviday, line 149: :+1:

corviday (approved):
LGTM. As far as I can tell, the only changes updating to new version of modelmeta and the database were the changes to requirements.txt, and everything else was docker and CI wrangling (aside from the metadata download fix) Or did I miss something?

Demo works great.

setup.py
```diff
@@ -52,51 +65,52 @@ def get_commitish():
...
-    command_options={
-        'build_sphinx': {
-            'build_dir': ('setup.py', 'pdp/docs')
-        }
+    install_requires=[
+        "flask",
+        "beaker",
+        "genshi",
+        "static",
+        "pdp-util",
```
jameshiebert, line 81: Should `pdp-util` have a hard requirement of `>=` some version?

jameshiebert (commented):
Most everything looks *great* to me. Thanks for solving the Python-CI issue!

setup.py
rod-glover: That's perhaps a bigger discussion than fits here. I've taken the discussion to our Zulip.

rod-glover:
Snyk is flagging several unfixable vulns, so I am going to merge anyway and deal with the vulns later.

PR #238, opened 2021-07-12
Organize Docs

### LINKED ISSUE #237: Documentation Organization
There is a lot of information both in the `README` and hidden throughout the `pdp`. All of the information is relevant and important but it can be challenging to find at times. We may want to consider consolidating all the `READMEs` into a `/docs` folder and adding something like a table of contents.

nikola-rados:
## Changes
- Moves `README` info to `DOCS`
- Combines other `READMEs` into one

I wanted to help organize all the helpful information we have in the `pdp`. I felt that we would benefit from moving most of it into a dedicated `/docs` directory. I am open to any suggestions on restructuring or reorganizing this, so please let me know!

Resolves #237 

## Checklist
- [ ] Full integration tests are passing locally


rod-glover (commented):
I like the terse front-page README.

Generally, this is what we discussed. I have the following thoughts:
1. Much of this info will be changed shortly. But this is a good placeholder.
2. I think it might be better to break up the big DOCS.md file into separate files and sections, say 
   - Dependencies + installation
   - Configuration + deployment
   - Testing
   - Module calendars
3. ... And either DOCS becomes like the README (short, with links) or you just hoist those subsections into the README ... I'd favour the latter.
4. Also, .md files don't have to have names in all caps. That's just a convention to make them easy to find in the top-level directory, I think.
5. The numbering is nice, but it looks like a bear to maintain ... unless you used some automated tool, which I'd be very interested to learn about.

docs/DOCS.md
```diff
@@ -0,0 +1,835 @@
...
+   
+1. Stop the container: When you no longer wish the container to be running,
+   enter `ctrl+D` or `exit` at the container command line prompt. 
+   
+   If you detached from the image, you can enter 
+   `./docker/local-pytest/down-backend.sh` to stop and remove the container.
+
+### 4.1.2 What
+
+The files in this directory allow you to build and run a test environment
```
corviday, line 238: This README is no longer in the same directory as the tests. May need to specify which directory here.

docs/DOCS.md
```diff
@@ -0,0 +1,835 @@
...
+    mapping user ids between a container and its run environment, not for the
+    fainthearted. 
+
+1. As noted above, running tests in the test container in read/write mode 
+leaves problematic pycache junk behind in the host filesystem. 
+This can be cleaned up by running `py3clean`.
+
+
+## 4.2 Node.js (tests of JS code)
+
```
corviday, line 359: There is another README covering some details on some of the tricky details about  _writing_ javascipt tests [here](https://github.com/pacificclimate/pdp/blob/e9216e79049ecb4c6a5808f75704037bcd35eadb/pdp/static/js/__test__/README.md). I don't know if you want to include it as well, or if you only want info on _running_ tests in the giant README.

docs/DOCS.md
```diff
@@ -0,0 +1,835 @@
...
+With Node.js installed (see above), you can install all the test framework dependencies
+as follows:
+
+```bash
+npm install
+```
+
+Notes:
+
```
corviday, line 108: Would these running-javascript-tests notes fit better in section 4.2 on running tests, rather than section 2.1 on installing things?

nikola-rados:
1. Ya I sort of assumed a lot of it would be out the window but at least the format is better.
2. Great idea, that would help alleviate some of the work the table of contents has to do.
3. Agreed, more links in the `README` would be better.
4. Understood.
5. I did it manually. It helped to break up the sections a bit but if we breakdown the readme into several files it should make it the numbers redundant.

docs/DOCS.md
```diff
@@ -0,0 +1,835 @@
...
+so we install `jest-environment-jsdom-fourteen`, which upgrades the version of
+`jsdom`. This may become unnecessary with later versions of Jest.
+
+* Since little of the JS code is written with unit testing in mind,
+we exploit `jsdom` heavily in the tests. Essentially, these tests use
+jQuery queries to find out what is going on in the DOM as the app does its thing.
+
+
+## 2.2 Production
+
```
corviday, line 124: Not sure what to do with this section. It's not what we do in production, production is all dockerized now. But it may be worth documenting installing via `pyenv` and `pip` as an alternative to `tox`. I think maybe tox doesn't work in python 3.x? I remember some sort of trouble along those lines recently.

I guess I'd split this into two sections. Move the non-tox install into section 2.1.1 as an alternative to tox, if people don't want to use it or it doesn't work. Make the how-to-build-sphinx documentation its own section, 2.1.3 Build Documentation.

Maybe this stuff doesn't matter if it is all going to be overhauled anyway.

docs/DOCS.md
```diff
@@ -0,0 +1,835 @@
...
+// Note these are `CfDatetime`s, not CalendarDatetime`s.
+const start = cfTimeSystem.firstCfDatetime();
+console.log(start.index);  // -> 0
+
+const end = cfTimeSystem.lastCfDatetime();
+console.log(end.index);  // -> 99999
+
+const today = cfTimeSystem.todayAsCfDatetime();
+console.log(today.index);  // -> some value around (<current year> - 1950) * 360
+```
```
corviday, line 835: This file seems too long to me, I can easily imagine myself getting lost looking for some specific thing. I would suggest splitting it up my topic or task, and having an intro / table of contents / FAQ / whatever that explains what is in each smaller help file and links to it.

nikola-rados:
@corviday and @rod-glover I have made some changes based on your feedback. I think we definitely need a content overhaul but since things are changing so soon we should try and avoid changing it twice. Furthermore, I've been looking into `pipenv` to replace some of the installation tools we have here (mainly `tox` since it never works) and that will have a big impact on the documentation.

corviday (approved):

rod-glover (approved):
This reorg is good. There's a lot of content updating to be done, but that's for later.

PR #249, opened 2021-09-19
Use bclite basemap

### LINKED ISSUE #244: Use "Lite" BC basemap: BC Prism portal

rod-glover:
## Changes
Use an OpenLayers `XYZ` layer pointed at new BC Lite basemap tiles

Resolves #244 

## Checklist
- [ ] Full integration tests are passing locally

### DIFF
--- a/docker/local-run/docker-compose.yaml
+++ b/docker/local-run/docker-compose.yaml
@@ -25,6 +25,7 @@ services:
       - NCWMS_URL=http://docker-dev02.pcic.uvic.ca:30779/dynamic/x
       - OLD_NCWMS_URL=https://services.pacificclimate.org/ncWMS-PCIC/wms
       - TILECACHE_URL=https://a.tile.pacificclimate.org/tilecache/tilecache.py https://b.tile.pacificclimate.org/tilecache/tilecache.py https://c.tile.pacificclimate.org/tilecache/tilecache.py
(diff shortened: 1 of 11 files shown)


rod-glover:
image
image
image

jameshiebert (commented):
The only questions/comments I have on this are with respect to the URL (http://142.104.230.53:30790/osm-bc-lite-test/$${z}/$${x}/$${y}.png).

1. Why the raw IP address? Can we get a DNS entry set up (if it doesn't exist already) and use that?). I realize that this is just a sample config, but it begs the question for whoever has to write the real config... where *should* this come from?
2. What's up with the double escaping in the XYZ? I.e. `$${z}/$${x}/$${y}.png` What should normally be in the URL and is that a function of the OpenLayers class that it is used in? If so, it might be a good idea to point to the OpenLayers docs, so the user has an idea of how to configure that URL (since our tilecache URLs use a different format).

Other that that, everything looks pretty straightforward.

rod-glover:
1. We now have a proxy for the tile service.
2. Double escaping is because environment variables are injected into JavaScript via a Python HTML templating tool, and that tool uses the same syntax for variable substitution as JavaScript. The `$$` is an escape for `$`, which results in the string `${z}/${x}/${y}.png` being injected. I will document this.

PR #253, opened 2021-10-01
Create a new portal for downscaled CMIP6 data

### LINKED ISSUE #250: Make portal to serve the CMIP6 data
Create backend and frontend support for a new portal to serve the BCCAQ2 downscaled CMIP6 data. Model it on the existing (CMIP5) Downscaled GCMS data, with any UI changes suggested by reviewers or scientists.

### LINKED ISSUE #252: logscale colour hack has stopped working with ncWMS upgrade
We want to display precipitation on the map with a logarithmic scale, otherwise the colour gradations are too subtle to see. This is difficult, because precipitation is often 0, which cannot be represented as a log of anything.

[ncWMS in theory provides a way to handle this](https://reading-escience-centre.gitbooks.io/ncwms-user-guide/content/04-usage.html#getmap): you can supply a range of values to be rendered with the COLORSCALERANGE` parameter, and than any values outside that range will be coloured as speified by the`BELOWMINCOLOR` and `ABOVEMAXCOLOR` parameters.

However, we found that the ancient version of ncWMS the PDP used to use (1.x running on pizza) ignored the `BELOWMINCOLOR` parameter, and rendered all values outside the range set by `COLORSCALERANGE` opaque black. As a workaround, we created a custom palette whose minimum value was black and whose maximum value was cyan, and used that to render logscaled precipitation, so that the minimum `COLORSCALERANGE`value, which was rendered as black on the custom palette, blended smoothly with the black outside-of-range values.

The PDP's ncWMS has been updated, and no longer renders all out-of-range values as black, so this strategy no longer works. Currently out-of-range values are rendered transparent, but hopefully we can now just use the `BELOWMINCOLOUR` parameter instead of complicated workarounds.

corviday:
This PR creates a new web portal so users can view and download CMIP6 datasets that have been downscaled with BCCAQ2. The portal is extremely similar to the Downscaled GCMS portal, where users can view and download CMIP5 BCCAQ2 datasets; they use the same all-Canada map component.

This PR also includes a minor update to the mapping code to take advantage of the updated ncWMS. Previously, with ncWMS 1, we could not use the `BELOWMINCOLOR` parameter to specify what colour zero precipitation should be rendered as when colours were assigned with a logarithmic scale. Now that @rod-glover has updated the PDP to use ncWMS 2, we are able to use this parameter to generate logscaled precipitation maps with no colour discontinuity.

Demo coming.

resolves #250 
resolves #252

### DIFF
--- a/pdp/static/js/gridded_observations_map.js
+++ b/pdp/static/js/gridded_observations_map.js
@@ -94,7 +94,7 @@ function init_obs_map() {
               min = 1;
               max = data.max;
               newParams["STYLES"] = "default/blueheat";
-              newParams["BELOWMINCOLOR"] = 'extend';
+              newParams["BELOWMINCOLOR"] = 'transparent';
             }
             else {
               newParams["LOGSCALE"] = "false";
--- a/pdp/static/js/gridded_observations_map.js
+++ b/pdp/static/js/gridded_observations_map.js
@@ -94,7 +94,7 @@ function init_obs_map() {
               min = 1;
               max = data.max;
               newParams["STYLES"] = "default/blueheat";
-              newParams["BELOWMINCOLOR"] = 'transparent';
+              newParams["BELOWMINCOLOR"] = 'extend';
             }
             else {
               newParams["LOGSCALE"] = "false";
--- a/pdp/static/js/gridded_observations_map.js
+++ b/pdp/static/js/gridded_observations_map.js
@@ -94,6 +94,7 @@ function init_obs_map() {
               min = 1;
               max = data.max;
               newParams["STYLES"] = "default/blueheat";
+              newParams["BELOWMINCOLOR"] = 'extend';
             }
             else {
               newParams["LOGSCALE"] = "false";
--- a/pdp/main.py
+++ b/pdp/main.py
@@ -18,6 +18,7 @@ import portals.bc_prism as bc_prism
 import portals.downscale_archive as downscale_archive
 import portals.bccaq2_downscale as bccaq2
 import portals.bccaq2_cmip6 as bccaq2_cmip6
+import portals.bccaq2_canesm5 as bccaq2_canesm5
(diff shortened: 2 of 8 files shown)


pdp/static/js/gridded_observations_map.js
```diff
@@ -94,6 +94,7 @@ function init_obs_map() {
               min = 1;
               max = data.max;
               newParams["STYLES"] = "default/blueheat";
+              newParams["BELOWMINCOLOR"] = 'extend';
```
rod-glover, line 97: Curious why this was updated too. `gridded_observations_map` doesn't seem to be part of this portal. Guessing it was just to spread the goodness while you were thinking of it.

rod-glover (approved):
The Python code could use a run through Black, but there's nothing wrong with it and that would only touch a small fraction of the codebase.

If the demo plays nicely, I'd say LGTM and go ahead!

corviday:
[Demo is up](https://services.pacificclimate.org/dev/portal/downscaled_cmip6/map/), seems good to me. We're awaiting any UI tweaks from RCI, then we'll merge. 

I'm not going to run black on this branch, since that would make rebasing the python3 branch onto it way more annoying.

corviday:
UI tweaks requested by RCI:

- [ ] revert logscaled precipitation colours back to transparent in cases of low or no precipitation
- [ ] format experiment strings like so: `Historical, SSP1-2.6`
- [ ] leave only the first CANESM5 run in the CMIP6 portal, create a separate page for the rest of them

corviday:
[This issue](https://github.com/pacificclimate/data-prep-actions/issues/52) tracks data processing for the new portal.

corviday:
[New CanESM5 portal page.](https://services.pacificclimate.org/dev/portal/downscaled_canesm5/map/)

I think the code is ready to go, except then RCI is going to think a little more about the page title. Datasets are still being processed.

PR #269, opened 2022-06-16
Add ability to import external documentation

### LINKED ISSUE #268: Add ability to import external documentation
Although it would be preferable to keep documentation with the application code it documents, i.e., in the linked repo, we wish to integrate the documentation of all data portals so that common information is included and the documentation UI (which, thanks to Sphinx, includes Search functions) is consistent and always up to date. Therefore the documentation will be maintained here.

rod-glover:
## Changes

The PDP will soon include some portals that are maintained as separate projects/repos (e.g., the [Met Data Portal](https://github.com/pacificclimate/station-data-portal)). We refer to these as "external" portals, to distinguish them from the ones implemented entirely inside this repo.

Add capability to download user documentation for external portals from GitHub and integrate it into the PDP user documentation suite.

Resolves #268

## Checklist


docker/production/Dockerfile
```diff
@@ -21,8 +24,11 @@ ADD --chown=${USERNAME}:${GROUPNAME} . ${USER_DIR}/
 RUN pip install -r requirements.txt -r deploy_requirements.txt
 
 # Install and build the docs
+# Must pre-install to provide dependencies and version number
```
jameshiebert, line 27: Thanks for documenting this so we don't accidentally lose our version number again.

setup.py
```diff
@@ -1,3 +1,29 @@
...
+    pip install .
+
+The first installation accomplishes two things:
+
+* make project version number available to Sphinx
+* install dependencies required by ``ExtendedBuildDoc``
+
+The second installation makes the docs available.
+
+These things could possibly be accomplished without a double installation,
```
jameshiebert, line 23: Pull Requests welcome :stuck_out_tongue:

pdp/doc_helpers/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    Download a file specified by a flattened manifest item,
+    storing it under the specified target directory (it may be in a subdir).
+
+    :param base_url: URL of directory containing files to download.
+    :param manifest_item: Flattened manifest item specifying file to download.
+    :param target_dir: Path specifying top-level target directory in which
+        to store file.
+    """
+    # Make target directory
+    item_dir = "/".join(manifest_item[:-1])
```
jameshiebert, line 113: Advisable to use `os.path.join` throughout, but I guess this branch is stuck in Python 2. This should be flagged for @eyvorchuk 's 2-to-3 conversion.

setup.py
```diff
@@ -1,3 +1,29 @@
+"""
+This is not your grannie's setup, mainly because of how we build documentation.
```
jameshiebert, line 2: :laughing:

pdp/doc_helpers/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+are represented by a list of file names; a (sub)directory is represented
+by a list item that is a dictionary whose key is the directory name and
+whose value is likewise a list of files or directories, recursively.
+
+For more details, see the README on this subject.
+"""
+from __future__ import print_function
+import subprocess
+from warnings import warn
+import urllib2
```
jameshiebert, line 31: You  may want to alert @eyvorchuk about the addition of this as it has implications for the 2-to-3 conversion that he'll need to work around.

docs/sep-proj-user-doc.md
```diff
@@ -0,0 +1,78 @@
+# External portals and user documentation
+
+As of the release that includes this README, the PDP will include some
+portals that are maintained as separate projects/repos (e.g., the
```
jameshiebert, line 4: Does this section get included in the user docs or is it just for devs? If it is only for devs, then just ignore the following...

I don't know that I agree with the way that this is communicated. I think that when we talk about a "portal" it's from the user/stakeholder perspective, whereas we talk about software/repositories from our own perspective as the service provider. This documentation is supposed to be user documentation, so I don't think that it's useful or even desirable to describe the nuances of what parts of the software are in which repositories or to differentiate between "external"/"internal". That's for us to worry about. The user just wants to know how to use it.

jameshiebert (commented):
Looks great, modulo a few small comments. Non-ideal to have to actually download from GitHub all of the external docs files, but as we discussed, it's probably the least-bad of the automatable options that we have. Your implementation of it looks great... thanks for doing that.

docs/sep-proj-user-doc.md
rod-glover: It's just for devs. Confusingly, we have 
- the `docs` directory for what amount to README extensions, for devs
- the `doc` directory containing user documentation

I hesitate to refactor this for fear of confusing Sphinx.

pdp/doc_helpers/__init__.py
rod-glover: We are indeed in Python 2.

rod-glover: No, wait, we have [`os.path.join` in Py 2.](https://docs.python.org/2.7/library/os.path.html#os.path.join)

rod-glover: Done

rod-glover:
Tested docs build with recent changes. All good.

jameshiebert:
:tada:

PR #278, opened 2022-09-29
I277 remove portals

### LINKED ISSUE #277: Remove Deprecated Raster Portals
Since the `downscaled_gcms_archive` portal is deprecated and the `bccaq_extremes` portal has never been used due to data problems (see also #124), these portals should be removed from the codebase.

eyvorchuk:
This PR resolves #277, which is to remove the `downscaled_gcms_archive` and `bccaq_extremes` raster portals due to them being deprecated.


doc/source/raster.rst
```diff
@@ -206,7 +206,7 @@ You can retrieve the catalog with your favorite programming languages as well. F
   > json_data[[1]]
   [1] "https://data.pacificclimate.org/data/downscaled_gcms/pr+tasmax+tasmin_day_BCCAQ+ANUSPLIN300+GFDL-ESM2G_historical+rcp45_r1i1p1_19500101-21001231.nc"
 
-At present, there are seven pages for which one can retrieve catalogs: ``bc_prism``, ``downscaled_gcms``, ``downscaled_gcms_archive``, ``downscaled_gcm_extremes``, ``gridded_observations``, ``hydro_model_archive``, and ``hydro_model_out``.
+At present, there are five pages for which one can retrieve catalogs: ``bc_prism``, ``downscaled_gcms``, ``gridded_observations``, ``hydro_model_archive``, and ``hydro_model_out``.
```
jameshiebert, line 209: It looks like we forgot to update this line when we put in the CMIP6 portals. Can you add `downscaled_canesm5` and `downscaled_cmip6` to the list?

jameshiebert (commented):
Looks good to me aside from one small addition. And nice catch on realizing that d3 isn't used anymore with the removal of the extremes portal. I would have missed that. Do you have a fresh run of the full tests (against gluster storage) to ensure that everything's working?

eyvorchuk:
> Do you have a fresh run of the full tests (against gluster storage) to ensure that everything's working?

Just ran the full test suite. They all work properly after making the slight modifications I just pushed.

PR #286, opened 2023-05-30
Remove BC Station Data Portal (PCDS) frontend

### LINKED ISSUE #284: Remove BC Station Data Portal (PCDS) frontend

rod-glover:
## Changes

- Remove PCDS frontend Python code. The backend code must remain in place. There is a very small finesse needed to eliminate the frontend creation.
- Remove PCDS frontend JS code -- all and only code used to implement PCDS

Resolves #284 

## Checklist
- [x] ~Docs have been updated to reflect the changes~ this is the subject of a different PR


rod-glover:
@eyvorchuk , you can confine yourself to looking at the Python code if you prefer.
@corviday , I've asked you particularly for your knowledge of the JS code. But feel free to comment on anything.

pdp/main.py
```diff
@@ -37,11 +37,11 @@ def initialize_frontend(global_config, use_analytics=False):
     static_app = static.Cling(resource_filename('pdp', 'static'))
 
     mounts = {
+        # Omit pcds frontend, because deprecated. (But we need the backend still.)
```
corviday, line 40: Why do we still need the backend? Does the new frontend use it?

pdp/static/js/__test__/canada_ex_app.test.js
```diff
@@ -26,7 +26,6 @@ require('./globals-helpers').importGlobals([
     { module: 'js/pdp_filters.js', spread: true },
     { module: 'js/pdp_map.js', spread: true },
     { module: 'js/pdp_raster_map.js', spread: true },
-    { module: 'js/pdp_vector_map.js', spread: true },
```
corviday, line 29: I see there are a bunch of test files where deleting this line is the only change, was it for something not actually used in the tests, and just present due to copy and paste test coding?

corviday (approved):
Looks good to me. I took a quick peak through the repository to see if I spotted anything you missed, but as far as I can tell, you got it all.

pdp/static/js/__test__/canada_ex_app.test.js
rod-glover: My post-hoc diagnosis is copy-paste, or otherwise mistaken inclusion in the first place, as you say.

pdp/main.py
rod-glover: It does. I don't want to remove it, although we do have a separate branch `pcds-only`, that is for the new frontend -- it's just the backend, as far as I know. James kindly built it for that purpose. In principle it _could_ be removed here since we are deploying for that purpose from the branch, not from the mainline. But that branch is intended eventually to be merged into the mainline, and I don't want to remove its basis before we do that.

PR #303, opened 2023-10-13
I298 tiles url

### LINKED ISSUE #298: Replace tilecache.py requests with new tiles URL
Currently, the `downscaled_*` and `gridded_observations` portals make requests to `tilecache.py` and should instead make requests to a different tiles URL given by https://services.pacificclimate.org/tiles/lite-wgs84. I believe this is a matter of replacing the usage of `getNaBaseLayer` in `canada_ex_map.js` and `gridded_observations_map.js` with `getTileBaseLayer`, and replacing the `TILECACHE_URL` environment variable with a `TILES_URL` variable.

eyvorchuk:
This PR resolves #298. This replaces the uses of Tilecache with PCIC's configuration of [MapProxy](https://mapproxy.org/), which serves basemap tiles from [OpenStreetMap](https://www.openstreetmap.org) in the North America portals.

Demo available [here](https://services.pacificclimate.org/dev/portal/downscaled_cmip6/map/).

## Checklist

### DIFF
--- a/pdp/__init__.py
+++ b/pdp/__init__.py
@@ -38,7 +38,7 @@ def get_config_from_environment():
         'geoserver_url': 'http://tools.pacificclimate.org/geoserver/',
         'ncwms_url': 'http://tools.pacificclimate.org/ncWMS-PCIC/wms',
         'old_ncwms_url': 'https://services.pacificclimate.org/ncWMS-PCIC/',
-        'tiles_url': 'https://tools.pacificclimate.org/tiles',
+        'na_tiles_url': 'https://tools.pacificclimate.org/tiles/lite-wgs84',
         'bc_basemap_url': 'http://142.104.230.53:30790/osm-bc-lite-test/$${z}/$${x}/$${y}.png',
         'use_analytics': 'True',
         'analytics': 'UA-20166041-3'  # change for production
--- a/pdp/__init__.py
+++ b/pdp/__init__.py
@@ -38,7 +38,7 @@ def get_config_from_environment():
         'geoserver_url': 'http://tools.pacificclimate.org/geoserver/',
         'ncwms_url': 'http://tools.pacificclimate.org/ncWMS-PCIC/wms',
         'old_ncwms_url': 'https://services.pacificclimate.org/ncWMS-PCIC/',
-        'tiles_url': 'https://tools.pacificclimate.org/tiles/lite-wgs84',
+        'wgs84_url': 'https://tools.pacificclimate.org/tiles/lite-wgs84',
         'bc_basemap_url': 'http://142.104.230.53:30790/osm-bc-lite-test/$${z}/$${x}/$${y}.png',
         'use_analytics': 'True',
         'analytics': 'UA-20166041-3'  # change for production
--- a/pdp/__init__.py
+++ b/pdp/__init__.py
@@ -38,7 +38,7 @@ def get_config_from_environment():
         'geoserver_url': 'http://tools.pacificclimate.org/geoserver/',
         'ncwms_url': 'http://tools.pacificclimate.org/ncWMS-PCIC/wms',
         'old_ncwms_url': 'https://services.pacificclimate.org/ncWMS-PCIC/',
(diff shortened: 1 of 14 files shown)


pdp/static/js/pdp_map.js
```diff
@@ -154,9 +154,11 @@ function getNaBaseLayer(wmsurl, displayname, layername, proj) {
         wmsurl,
         {
             layers: layername,
+            transparent: "true"
```
QSparks, line 157: There it is. Nice job!

pdp/static/js/canada_ex_map.js
```diff
@@ -1,5 +1,5 @@
 /*jslint browser: true, devel: true */
-/*global $, jQuery, OpenLayers, pdp, map, na4326_map_options, getBasicControls, getBoxLayer, getEditingToolbar, getHandNav, getBoxEditor, getNaBaseLayer, getOpacitySlider, Colorbar*/
+/*global $, jQuery, OpenLayers, pdp, map, na4326_map_options, getBasicControls, getBoxLayer, getEditingToolbar, getHandNav, getBoxEditor, getNaBaseLayer, getTileBaseLayer, getNa4326LiteBaseLayer, getOpacitySlider, Colorbar*/
```
QSparks, line 2: Is getTileBaseLayer used anymore?

QSparks (commented):
LGTM! I added a couple of questions and have a suggestion that we update some function names to better reflect the new OSM tiles, such as changing `getNa4326LiteBaseLayer` to `getOSM4326BaseLayer`.

Well done overall!

pdp/__init__.py
```diff
@@ -38,10 +38,7 @@ def get_config_from_environment():
         'geoserver_url': 'http://tools.pacificclimate.org/geoserver/',
         'ncwms_url': 'http://tools.pacificclimate.org/ncWMS-PCIC/wms',
         'old_ncwms_url': 'https://services.pacificclimate.org/ncWMS-PCIC/',
-        'tilecache_url':
-            'http://a.tiles.pacificclimate.org/tilecache/tilecache.py'
-            ' http://b.tiles.pacificclimate.org/tilecache/tilecache.py'
-            ' http://c.tiles.pacificclimate.org/tilecache/tilecache.py',
+        'na_tiles_url': 'https://tools.pacificclimate.org/mapproxy/service',
```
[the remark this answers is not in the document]
eyvorchuk: I was under the impression that we just use `tools.pacificclimate.org` as a placeholder URL for default values, but according to [this Confluence page](https://pcic.uvic.ca/confluence/display/CSG/Cloud_Present), it looks like a proxy URL for pizza. So, yeah, I'll change this to `services.pacificclimate.org`.

pdp/static/js/canada_ex_map.js
eyvorchuk: No, and neither is `getNa4326LiteBaseLayer`. I was thinking about keeping these in just in case we decide to use XYZ instead of WMS to get layers later down the line, but I'll see how others feel.

QSparks: Gotcha, I think you’re right to keep it just in case.

QSparks (approved):
LGTM!

jameshiebert (commented):
Looks good @eyvorchuk and @QSparks ! The code changes are pretty straightforward, since most of the behaviour details are hidden in the mapproxy setup. I think that you both came up with a creative solution to solve the problem expediently.

I do have one critique (and to be very clear, I am *not* asking you to change what you've done). I think that using a single land mask ends up being non-optimal for a couple reasons. First of all it's not scale-adaptive, like the pre-generated tiles are. I.e. with the tiles, you get more detail in the shoreline as you move up in zoom levels. Whereas having a single land mask, the shoreline has awkward aliasing at low zoom levels and is a little too coarse at high zoom levels.

image

Secondly, labels of coastal areas have a tendency to get cut off.

image

Again, please don't spend time on trying to fix these. It will get worked out if/when we move to a better long term solution of vector tiles. In the mean time, this achieves the short-term goals of decommissioning the tile server and replacing it with a reasonable map.

Issue #320, opened 2025-03-20
Ubuntu runner and node updates


QSparks:
Use Ubuntu-24.04 runners and node 22

PR #321, opened 2025-03-21
I319 and i320

### LINKED ISSUE #319: Update default dataset to PCIC Blend, Tasmax
The PNWNAmet 2015 Wind variable is the current default dataset displayed when starting the [Daily Gridded Meteorological Datasets](https://data.pacificclimate.org/portal/gridded_observations/map/) portal. 

It is the only variable that is not clipped to the land, which causes it to appear slightly offset compared to the other datasets and doesn’t present as well visually.
I've proposed an update to Markus and he has suggested PCIC Blend, Tasmax as the starting variable. 

A quick search of the repo suggests we would want to update this variable:
https://github.com/pacificclimate/pdp/blob/702966337cb850f6ec572d66ca88f3c5c47253c4/pdp/static/js/gridded_observations_map.js#L25

to use `tasmax_day_PCIC_BLEND_v1_historical_19500101-20121231_Canada` instead.

QSparks:
## Changes
-  9482a39  `PCIC-Blend Tasmax` the default dataset for the [Gridded Observations Portal](https://beehive.pacificclimate.org/orca-pdp/portal/gridded_observations/map/)
- Use node `22` and Ubuntu `24.04` runners.

Resolves #319 and #320

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -3,14 +3,13 @@ name: Docker Publishing
 on:
   push:
     branches:
-      - '*'
+      - "*"
     tags:
-      - '[0-9]+.[0-9]+.[0-9]+'
+      - "[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
   publish:
-
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
       - uses: actions/checkout@master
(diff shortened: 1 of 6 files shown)


corviday (approved):
LGTM!

What was the rationale behind changing the initial dataset?

QSparks:
> What was the rationale behind changing the initial dataset?

The wind variable is the only one not clipped to the land, which causes it to appear slightly offset compared to the other datasets and, in my opinion, doesn’t look as good.

PR #322, opened 2025-08-01
Include .git for revision SHA at docker build time

### LINKED ISSUE #242: Data portal version string broken
The footer now reads "PCIC Data Portal version unknown (unknown)". Version information has become unavailable as of version 4.0.0.

This information is extracted in [`pdp/__init__.py`](https://github.com/pacificclimate/pdp/blob/758b8e76cacfeefef441aa44431eec3acf7a7730/pdp/__init__.py#L97-L98) using `pkg_resources`. 

A possible cause is the rearrangement of Docker images in this release, specifically including change of USER inside images, making some installation information unavailable.

QSparks:
## Changes
- Remove .git from .dockerignore
- Remove .git after docker build

Versions now show up correctly:
`PCIC Data Portal version 4.4.1 (i242.fix.version.filling:fb956e)`

Demo on orca-pdp:
https://beehive.pacificclimate.org/orca-pdp/portal/gridded_observations/map/

Resolves #242

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,7 +2,6 @@
 .bash_history
 
 # Git
-.git
 .github
 .gitignore
 
--- a/docker/production/Dockerfile
+++ b/docker/production/Dockerfile
@@ -20,6 +20,7 @@ USER ${USERNAME}
 
 ADD --chown=${USERNAME}:${GROUPNAME} . ${USER_DIR}/
 
+
 # Install dependencies. Note: Base image already contains several of the
 # heaviest ones.
 RUN pip3 install -r requirements.txt -r deploy_requirements.txt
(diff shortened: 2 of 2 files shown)


corviday (approved):
Nice detective work! Makes sense, and I have verified that the demo now displays a version.

Nospamas:
I think that you include .git at any point it'll be included as a layer within the image and therefore you won't get the space saving benefit of not including it in the first place. Doesn't look like this effects image size too much however: [historic image sizes](https://hub.docker.com/r/pcic/pdp/tags) and [where this image's size comes from](https://hub.docker.com/layers/pcic/pdp/i242-fix-version-filling/images/sha256-ede2d61e79a2d8f57cec7a023da705d521b80eb73783c88f67afd6d26fb434ba)

Though 500MB images are a little concerning in themselves, we'll have to take that into another issue 🛩️

Nospamas (approved):
LGTM

